PR #19, opened 2021-07-02
Pip-489 - Initialize e2e suite

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-489

### Change description ###
- Initialize e2e test with puppeteer
- add `yarn test:e2e` command to the scripts lists in `package.json`
- update README instructions to run e2e
- add 2 basic validation steps to verify start page is rendered and search option page is rendered after Start now button click

```
[ ] Yes
[x] No
```


src/test/e2e/e2e.ts
```diff
@@ -0,0 +1,43 @@
+import { Browser, Page } from 'puppeteer';
+
+const puppeteer = require('puppeteer');
+
+const port: number = parseInt(process.env.PORT, 10) || 8080;
+
+describe('End to end test', () => {
```
kurt-uk, line 7: Can we try a Page Object Model approach to our e2e? i think using a common testing approach like that would be 1) easier to read 2) easier to maintain

kurt-uk: the describe should be the journey we are completing so something closer to `finding a court or tribunal listing`

src/test/e2e/e2e.ts
```diff
@@ -0,0 +1,43 @@
...
+  const PAGE_URL = `https://localhost:${port}`;
+  let browser: Browser;
+  let page: Page;
+
+
+  beforeAll(async () => {
+    browser = await puppeteer.launch({
```
kurt-uk, line 14: all of this should belong in the config not the actual test file

src/test/e2e/e2e.ts
```diff
@@ -0,0 +1,43 @@
...
+      headless: false, // set to false and uncomment slowMo and args to see tests in realtime
+      slowMo: 80,
+      args: ['--window-size=1920,1080'],
+      ignoreHTTPSErrors: true, // skips 'Your connection is not private' page
+    });
+  });
+
+  test('should open main page with "Find a court or tribunal listing" title', async () => {
```
kurt-uk, line 22: use `it` instead of `test`

kurt-uk: if we have pupeteer as async, should we try use lamda functions instead?

kurt-uk (changes requested):

src/test/e2e/e2e.ts
kurt-uk: > Can we try a Page Object Model approach to our e2e? i think using a common testing approach like that would be 1) easier to read 2) easier to maintain

or a Screenplay pattern? both would be useful

src/test/e2e/PageObjects/HomePage.ts
```diff
@@ -0,0 +1,31 @@
+import { Page } from 'puppeteer';
+
+const helpers = require('../Helpers/Selectors');
+
+class HomePage {
+  async OpenHomePage(page: Page): Promise<any> {
+    await page.goto(helpers.baseUrl, {waitUntil: 'domcontentloaded'});
+
+    const pageTitle = await page.$eval(helpers.MainHeader, (e: Element) => e.textContent);
+    expect(pageTitle).toBe('Find a court or tribunal listing');
```
kurt-uk, line 10: we should have the tests on the test page, these page objects should return pages we are on and the evaluating be done on the object

src/test/e2e/PageObjects/HomePage.ts
```diff
@@ -0,0 +1,31 @@
+import { Page } from 'puppeteer';
```
kurt-uk, line 1: can we tag our page object files as `blahblah.po.ts` even though they're in the page object folder youve created, just nice to have them tagged as .po

src/test/e2e/PageObjects/HomePage.ts
```diff
@@ -0,0 +1,31 @@
...
+    await page.click(helpers.StartNowButton);
+    await page.waitForSelector(helpers.SearchOptionsTitle).catch(() => {
+      console.log(`${helpers.SearchOptionsTitle} not found`);
+    });
+
+    const pageTitle = await page.$eval(helpers.SearchOptionsTitle, (e: Element) => e.textContent);
+    const radioElements = await page.$$(helpers.RadioButton);
+
+    expect(pageTitle).toContain('Find a court or tribunal list');
+    expect(radioElements.length).toBe(2);
```
kurt-uk, line 27: this isnt the home page, if we return pages instead of expected it would make it easier to read and follow

src/test/e2e/Tests/e2e.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const puppeteer = require('puppeteer');
+const HomePage = require('../PageObjects/HomePage');
+const homePage = new HomePage;
+let browser: Browser;
+let page: Page;
+
+beforeAll(async () => {
+  browser = await puppeteer.launch({
+    headless: false, // set to false and uncomment slowMo and args to see tests in realtime
```
kurt-uk, line 11: this is still a lot of config in the test suite, this needs moving to a puppeteer config file somewhere

src/test/e2e/Tests/e2e.ts
```diff
@@ -0,0 +1,32 @@
+import { Browser, Page } from 'puppeteer';
+
+const puppeteer = require('puppeteer');
+const HomePage = require('../PageObjects/HomePage');
+const homePage = new HomePage;
+let browser: Browser;
+let page: Page;
```
kurt-uk, line 7: if we return our own pages we can use those and the standard browser object in ours instead of the browsers new pages

marcogagliardi (approved):
👍

kurt-uk (approved):

PR #20, opened 2021-07-05
pip-490: css for small screen

marcogagliardi:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-490

### Change description ###
reformat header
css for small screen
added sign-in
Screenshot 2021-07-05 at 14 17 10

```

[ ] Yes
[X] No
```

### DIFF
--- a/src/main/assets/scss/main.scss
+++ b/src/main/assets/scss/main.scss
@@ -10,10 +10,6 @@
   }
 }
 
-.govuk-header__product-name {
-  font-weight: 800;
-}
-
 
 
 
--- a/src/main/assets/scss/main.scss
+++ b/src/main/assets/scss/main.scss
@@ -1,7 +1,8 @@
 @import '~govuk-frontend';
 
 .govuk-header__navigation--end {
-  margin: -3vh 0 0 0;
+  margin: -5vh 0 0 0;
+  padding: 1vh 0 0 0;
 }
 
 
--- a/src/main/assets/scss/main.scss
+++ b/src/main/assets/scss/main.scss
@@ -1,13 +1,7 @@
 @import '~govuk-frontend';
 
-@media (min-width: 48.0625em) {
-  .govuk-header__content {
-    width: 39%;
-  }
-
-  .govuk-header__logo {
-     width: 60%;
-  }
+.nav-right {
+  margin: -4vh 0 0 0;
 }
 
 
--- a/src/main/assets/scss/main.scss
+++ b/src/main/assets/scss/main.scss
@@ -1,7 +1,7 @@
 @import '~govuk-frontend';
 
-.nav-right {
-  margin: -4vh 0 0 0;
+.govuk-header__navigation--end {
+  margin: -3vh 0 0 0;
 }
 
 
--- a/src/main/assets/scss/main.scss
+++ b/src/main/assets/scss/main.scss
@@ -1,14 +1,20 @@
 @import '~govuk-frontend';
 
-.nav-right {
-    float: right;
-    margin: -1.6vw 0px 0vh 0;
-    li {
-      a {
-        font-size: 19px;
-      }
-    }
+@media (min-width: 48.0625em) {
+  .govuk-header__content {
+    width: 39%;
+  }
+
+  .govuk-header__logo {
+     width: 60%;
+  }
 }
+
+.govuk-header__product-name {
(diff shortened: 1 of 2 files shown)


src/main/views/template.njk
```diff
@@ -11,12 +11,12 @@
 
 {{ govukHeader({
         homepageUrl: "#",
-        serviceName: "Find a court or tribunal listing",
+        productName: "Find a court or tribunal listing",
```
kurt-uk, line 14: should this not be service name still? its moved it right across to the left now

src/main/assets/scss/main.scss
```diff
@@ -1,14 +1,20 @@
...
-        font-size: 19px;
-      }
-    }
+@media (min-width: 48.0625em) {
+  .govuk-header__content {
+    width: 39%;
+  }
+
+  .govuk-header__logo {
+     width: 60%;
```
kurt-uk, line 9: im not sure changing too much of the gov stylings is the best way to go, whats the logic for going against the gov styling here?

src/main/views/template.njk
marcogagliardi: N. It has been decided that it is product name and the prototype has been changed accordingly. Please see prototype.

src/main/assets/scss/main.scss
marcogagliardi: it is because the title is long and it is displayed in 2 lines at the bottom of the logo instead of one line at the top aking the header too high.

kurt-uk:
Screenshot 2021-07-05 at 15 12 07

surely this cant be what we're after @marcogagliardi ?

src/main/views/template.njk
kurt-uk: just seen the prototype but it was changed though your suggestion?

marcogagliardi: sorry what do you mean?

marcogagliardi:
> Screenshot 2021-07-05 at 15 12 07
> 
> surely this cant be what we're after @marcogagliardi ?

you haven't rebuild properly the project @kurt-uk. you need to run 'yarn webpack'. Do it and you will see.

src/main/views/template.njk
kurt-uk: well if youre steering the design then saying "look at the design" doesnt mean that its the right way

marcogagliardi: @kurt-uk I don't know what you want to say with this. I haven't driven the design. This was decided by @johnoneill46 with Junior and they just send me a live example in the channel https://hmcts-reform.slack.com/archives/C027779QH40 where you are a member too. If you are not happy with the design speak to John and Junior then. Here I am just developing what they decide. I can raise my concerns but that's it. I am not the decision maker mate.

kurt-uk: Here im only talking about the `productName` vs the `serviceName` to which id been told that it was changed to match the code due to navigation concerns, i have no issues with the design and i think you're getting off topic a little. i asked you a question at the beginning because i was curious as to why we had changed from the recommended attribute
Screenshot 2021-07-05 at 17 06 15

marcogagliardi: so are you happy to resolve this or do you need an action from me?

kurt-uk: could you revert to `serviceName` please? once its back the the original design itll be all ready to be tested 👍🏼

marcogagliardi: code reverted

src/main/views/template.njk
```diff
@@ -13,10 +13,10 @@
         homepageUrl: "#",
         serviceName: "Find a court or tribunal listing",
         serviceUrl: "#",
-        navigationClasses: 'nav-right',
+        navigationClasses: 'govuk-header__navigation--end nav-right',
```
kurt-uk, line 16: ooo good find! youve left in `nav-right` though, may as well just extend the gov one to have the -4vh? ill approve either way, let you decide if you want 2 classes or not

kurt-uk (approved):

src/main/views/template.njk
marcogagliardi: amended css to overwrite gov.uk css to use one css class

kurt-uk:
nice work!

commitBlob (approved):

PR #32, opened 2021-07-30
Pip 506 Search Page

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-506

### Change description ###
- Initialize Search page
- create reusable continue and back macros
- hook up radio button 1 to redirect to the search page

```
[ ] Yes
[x] No
```


src/main/views/macros/form.njk
```diff
@@ -0,0 +1,14 @@
+{% macro submitButton(label = 'Continue', id = '', ariaLabel = '') %}
```
kurt-uk, line 1: should we name this file something other than `form`? it doesnt reflect what is in it, theres no form in here, maybe `common-components` or something along those lines?

kurt-uk (changes requested):
dont forget to update all tests, you need to add to the search-option tests youve extended, e2e tests and route test

src/main/views/macros/form.njk
```diff
@@ -0,0 +1,14 @@
+{% macro submitButton(label = 'Continue', id = '', ariaLabel = '') %}
+    {% from "govuk/components/button/macro.njk" import govukButton %}
+    {{ govukButton({
+        text: label
+    })}}
+{% endmacro %}
+
+{% macro goBack(path = '#', text = 'Back', id = '', ariaLabel = '') %}
+    {% from "govuk/components/back-link/macro.njk" import govukBackLink %}
```
kurt-uk, line 9: can we move the imports to the top too pls?

src/main/service/inputFilterService.ts
```diff
@@ -19,11 +19,11 @@ export class InputFilterService {
   }
 
   private checkNotNullOrEmpty(value): boolean {
-    return value != undefined || value != '';
+    return value != undefined && value != '';
```
commitBlob, line 22: Is this 'weak' comparison?

kurt-uk:
Screenshot 2021-08-09 at 13 50 41

src/main/views/hearing-list.njk
```diff
@@ -8,8 +8,11 @@
 
 {% block beforeContent %}
   {{ govukBackLink({
-    text: "Back",
-    href: referringPage
+    text: text,
```
commitBlob, line 11: Should we use goBack from common-components for this as well? Or we should replace goBack with back link from `govuk` library?

src/main/modules/helmet/index.ts
```diff
@@ -31,7 +31,7 @@ export class Helmet {
           fontSrc: [self, 'data:'],
           imgSrc: [self, googleAnalyticsDomain],
           objectSrc: [self],
-          scriptSrc: [self, googleAnalyticsDomain, "'sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU='"],
+          scriptSrc: [self, googleAnalyticsDomain, "'unsafe-inline'", "'unsafe-eval'"],
```
ChrisS1512, line 34: Rather than the unsafe options, I think we should move the inline scripts into their own files so that they are hosted on the same domain (and therefore won't get marked by the CSP policy)

src/main/assets/scss/main.scss
```diff
@@ -1,4 +1,5 @@
 @import '~govuk-frontend';
+@import 'accessible-autocomplete.min';
```
ChrisS1512, line 2: This seemed to be causing an error for me - it couldn't find the accessible-autocomplete.min file.

~accessible-autocomplete seemed to work though. Not sure what could be causing it?

src/main/views/hearing-list.njk
kurt-uk: i think just in here as this page can be arrived from 2 different routes, it may be more appropiate to go back to which ever side they came from by getting history rather than manually routing them backwards like we have done on the others

kurt-uk: becuase this page can be landed on from the search route or the alphabet route, so adding a href of  '/search-result' wouldnt fit if they came from alphabet side

src/main/modules/helmet/index.ts
kurt-uk: this will be handled in https://tools.hmcts.net/jira/browse/PUB-625

ChrisS1512 (approved):

kurt-uk (approved):
👍🏼

PR #35, opened 2021-08-02
PUB-571 - List View

ChrisS1512:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-506

### Change description ###

This adds the list view to the UI, accessible on /list?courtId=<id>

Two things remain on this:

1) To link up via the Search pages / work out whether this is captured by another ticket
2) To determine whether GET / Query Params is OK

```
[ ] Yes
[x] No
```
Screenshot 2021-08-03 at 11 45 15


src/main/assets/scss/main.scss
```diff
@@ -1,10 +1,35 @@
 @import '~govuk-frontend';
 
+$line-colour: #1d70b8;
+
 .govuk-header__navigation--end {
   margin: -5vh 0 0;
   padding: 1vh 0 0;
 }
 
+.line-bottom {
```
kurt-uk, line 10: we should be using the gov classes and extending those if we need to, check out https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/ , theres loads there that arent on the examples on toolkit and this is the stylesheet we use

src/main/views/list.njk
```diff
@@ -0,0 +1,77 @@
+{% extends "template.njk" %}
```
kurt-uk, line 1: should we rename the file to something more understandable than `list`?

src/main/controllers/ListController.ts
```diff
@@ -0,0 +1,32 @@
+import { Request, Response } from 'express';
```
kurt-uk, line 1: should we rename from `listController`? its quite vague as a list

src/main/views/list.njk
```diff
@@ -0,0 +1,77 @@
...
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+  {{ govukBackLink({
+    text: "Back",
+    href: "/"
+  }) }}
+{% endblock %}
+
+{% block content %}
+<h2 class="govuk-heading-l govuk-!-margin-bottom-8">{{courtName}} hearing list</h2>
+{% from "govuk/components/table/macro.njk" import govukTable %}
```
kurt-uk, line 16: whack those nunjuck imports to the top of the file 😎

kurt-uk (changes requested):
awesome! can you also add to `test/routes`

src/main/views/list.njk
kurt-uk: btw i dont mean the line ive commented on, i mean the name of this actual file

kurt-uk:
also we already have mocked data which can be accessed in `resources/actions`

src/main/views/list.njk
ChrisS1512: Moved to top of file

src/main/controllers/ListController.ts
ChrisS1512: Updated to hearingListController :)

src/main/views/list.njk
ChrisS1512: Updated to hearingList :)

src/main/assets/scss/main.scss
ChrisS1512: I've updated the <hr> to use the Gov style class, and extended that instead

src/main/assets/scss/main.scss
```diff
@@ -1,10 +1,34 @@
...
 
+.govuk-section-break--visible {
+  border-color: $line-colour;
+}
+
+.table-header {
+  border-spacing: 100px;
+}
+
+.govuk-table {
```
kurt-uk, line 18: do we need this here? are we not using standard gov uk table layout?

src/main/app.ts
```diff
@@ -30,7 +29,6 @@ new Nunjucks(developmentMode).enableFor(app);
 new Helmet(config.get('security')).enableFor(app);
 new Container().enableFor(app);
 
-app.use(favicon(path.join(__dirname, '/public/assets/images/favicon.ico')));
```
kurt-uk, line 33: no favicon?

src/main/assets/scss/main.scss
ChrisS1512: I've removed the the wrapping / spacing, however left the table-layout in.

The fixed width layout makes sure the columns are equal in size

src/main/app.ts
ChrisS1512: Good spot. I've added this back in

src/main/assets/scss/main.scss
kurt-uk: ah ok, have you seen `govuk-!-width-two-thirds` and the like? if so then all good, but they also have spacing classes, under their `width` folder

src/main/routes/routes.ts
```diff
@@ -6,6 +6,7 @@ export default function(app: Application): void {
 
   app.get('/', app.locals.container.cradle.homeController.get);
   app.get('/search-option', app.locals.container.cradle.searchOptionController.get);
+  app.get('/list', app.locals.container.cradle.hearingListController.get);
```
kurt-uk, line 9: the path for this page is still `list` do we match it to `hearing-list`?

ChrisS1512: Good spot - updated to hearing-list

src/test/unit/views/hearing-list.ts
```diff
@@ -0,0 +1,75 @@
...
+
+describe('List page', () => {
+  beforeAll(async () => {
+    await request(app).get(PAGE_URL).then(res => {
+      htmlRes = new DOMParser().parseFromString(res.text, 'text/html');
+    });
+  });
+
+  it('Check that back button exists', () => {
+    const header = htmlRes.getElementsByClassName('govuk-back-link');
```
kurt-uk, line 19: all your const's are `header` but this one is back button

src/test/unit/views/hearing-list.ts
```diff
@@ -0,0 +1,75 @@
...
+  });
+
+  it('Should display table caption', () => {
+    const header = htmlRes.getElementsByClassName('govuk-table__caption--m');
+    const formattedDate = moment().format('MMMM DD YYYY');
+    expect(header[0].innerHTML).contains('List for ' + formattedDate, 'Could not find the table header');
+  });
+
+  it('Should contain expected column headings', () => {
+    const header = htmlRes.getElementsByClassName('govuk-table__header');
```
kurt-uk, line 36: this one is the table

src/main/views/hearing-list.njk
```diff
@@ -0,0 +1,77 @@
...
+
+{% extends "template.njk" %}
+{% block pageTitle %}
+    {{ "Hearings List" | safe }}
+{% endblock %}
+
+{% block beforeContent %}
+  {{ govukBackLink({
+    text: "Back",
+    href: "/"
```
kurt-uk, line 12: your back link goes all the way home, it should go back a step

kurt-uk: should just go back

src/main/assets/scss/main.scss
ChrisS1512: Ah, yeah I have indeed. I did actually play around with those originally, however there isn't one unfortunately for one-sixth, which is the one we would need for our table.

kurt-uk: 👍🏼  nice job checking it all out :)

src/test/unit/views/hearing-list.ts
ChrisS1512: Good spot :) I've updated these now

src/test/unit/views/hearing-list.ts
ChrisS1512: Updated :)

src/test/unit/controllers/HearingListController.ts
```diff
@@ -0,0 +1,52 @@
+import HearingListController from "../../../main/controllers/HearingListController";
+import sinon from 'sinon';
+import { Request, Response } from 'express';
+
+describe('Hearing list Controller', () => {
+  test("Check that court ID that exists renders the list page", () =>  {
```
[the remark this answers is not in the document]
ChrisS1512: Updated :)

src/main/views/hearing-list.njk
ChrisS1512: Good spot. I've changed this so that it uses the referrer passed through via the controller.

I tried it out by redirecting the home page to here locally and seemed to work OK

src/test/unit/views/hearing-list.ts
```diff
@@ -0,0 +1,75 @@
...
+import request from 'supertest';
+import moment from 'moment';
+
+import { app } from '../../../main/app';
+
+const PAGE_URL = '/hearing-list?courtId=2';
+
+let htmlRes: Document;
+
+describe('List page', () => {
```
kurt-uk, line 11:
```suggestion
describe('Hearing List page', () => {
```

kurt-uk (approved):

kurt-uk (approved):

kurt-uk (approved):

johnoneill46 (approved):

PR #48, opened 2021-08-17
PUB-507 - Fix for IE11 + Styling Changes

ChrisS1512:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-507

### Change description ###

This PR fixes the following:

- Back to Top was not working on IE11, due to InteractionObserver not being present. Code added to revert to using scrolling, when InteractionObserver is not supported
- Fixed issue where the link styling
- Fixed issue with positioning of back to top button

Also, moved to using Typescript for Back to Top functionality, and updated webpack accordingly.

```
[ ] Yes
[x] No
```


src/main/alphabetical.ts
```diff
@@ -0,0 +1,67 @@
+
```
kurt-uk, line 1: can we add this to a deeper directory? into assets/js or something

tsconfig.json
```diff
@@ -4,13 +4,14 @@
     "module": "commonjs",
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
-    "target": "es6",
+    "target": "es5",
```
kurt-uk, line 7: whats the reasoning for this? es6 would be best would it not?

kurt-uk (changes requested):

src/main/alphabetical.ts
ChrisS1512: good spot. I've moved both this and the other webpack JS file (index.js) that creates the bundles into a seperate folder

tsconfig.json
ChrisS1512: It would :( However unfortunately IE11 only supports ES5, therefore needed to lower it

kurt-uk: hmm that brings us 6 years out of date, but fair enough if thats the only option

commitBlob: There are pollyfills which transpile es6 code into vanilla javascript, rather than downgrading maybe we should try to use them?
Might be the matter of import only? With Angular and React there is usually pollyfills file and you just need to uncomment lines to enable it
https://polyfill.io/v3/url-builder/

commitBlob: https://github.com/commitBlob/folioAngular/blob/master/src/polyfills.ts
This is Angular example

kurt-uk: i saw something similar, this is another option https://babeljs.io/

ChrisS1512: Ah nice! I'll have a look at those now

commitBlob (approved):

src/main/bundles/alphabetical.ts
```diff
@@ -0,0 +1,67 @@
+
```
kurt-uk, line 1: would this file not be better kept with the name `back-to-top`? then if we need another back to top we can re-use?

ChrisS1512: I've now moved the BackToTop class to a separate file, which is now called from the alphabetical.ts one (this one still needs to be there as it's an entrypoint for webpack)

src/main/views/alphabetical-search.njk
```diff
@@ -74,5 +75,5 @@
 
 {% endblock %}
 {% block bodyEnd %}
-  <script src="assets/js/back-to-top.js"></script>
+  <script src="alphabetical-dev.js"></script>
```
kurt-uk, line 78: did you need to change it here too or is this something else?

ChrisS1512: nah this one needs to stay as this (the frontend loads the bundle, which loads the back-to-top)

kurt-uk: ok brill, as long as everything runs 👍🏼

kurt-uk (approved):

PR #50, opened 2021-08-18
PUB-569 search validation

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-569

### Change description ###
- prevent form resubmit on refresh, back and forward browser clicks
- set autocomplete to overlay to avoid focus issues
- link errors in the error summary to the answer they relate to

```
[ ] Yes
[x] No
```


src/main/views/search.njk
```diff
@@ -20,7 +20,7 @@
                     errorList: [
                         {
                             text: errorMessage,
-                            href: "#"
+                            href: "#search-input-container"
```
kurt-uk, line 23: the anchor tag here doesnt return focus to the input from when i tried it, not sure if youre using the right ref, also if you are using the right ref, can we rename? having `.../#search-input-container` in the url doesnt look fab

src/main/views/search.njk
```diff
@@ -70,12 +70,21 @@
...
           element: element,
           id: 'search-input',
-          source: source
-        })
+          source: source,
+          displayMenu: 'overlay'
+        });
+      history.pushState(null, null, location.href);
+      history.back();
+      history.forward();
```
kurt-uk, line 83: this seems extremely hacky? forcing them back and forward?

kurt-uk (changes requested):

src/main/views/search.njk
commitBlob: This was the only solution I found that covered back, forward, and refresh cases.
Others only worked for back and refresh

src/main/views/search.njk
commitBlob: It doesn't focus it jumps to that element on the page.

src/main/views/search.njk
```diff
@@ -45,7 +45,7 @@
                 {% if noResultsError or invalidInputError %}
                     <span class="govuk-error-message">Enter a court name, town or city</span>
                 {% endif %}
-                <div class="{{ inputClasses }}" id="search-input-container"></div>
+                <div class="{{ inputClasses }}" id="search-input"></div>
```
kurt-uk, line 48: we have 2 elements with the same id now, youll have to either change the name of this again or change the id in the autcomplete

ChrisS1512:
I'm getting an issue where the browsers back button isn't working in the following scenario:

1) Navigate to page and search for something that causes an error
2) Click the link in the error text
3) Click the browsers back button, which doesn't then take you back

ChrisS1512 (approved):

kurt-uk (approved):

PR #51, opened 2021-08-20
fix - use redirect instead of render to prevent form resubmission error

commitBlob:
### Change description ###
Use redirect instead of render on the POST requests in order to prevent form resubmission (PRG pattern)

```
[ ] Yes
[x] No
```


ChrisS1512 (approved):

kurt-uk (approved):
whislt this does work with the form resubmission error, the browser back button registers hitting this as a valid history visit, so clicking back doesnt seem to take you anywhere

PR #78, opened 2021-09-17
pub-720:  Add Live Status Hearing Updated View

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-720

### Change description ###
Add Live Status hearing updates page which builds on PUB-514. If there are live hearings for a court table is rendered, otherwise error message is shown. Page also has a timer to auto refresh every x seconds depending on the environment variable. If variable is not set timer defaults to 15 seconds.
In agreement with Junaid who is working on 721 links in the table point to `status-description` page which will be implemented later.

```
[ ] Yes
[x] No
```
Capture


kurt-uk (changes requested):

src/main/controllers/LiveStatusController.ts
```diff
@@ -0,0 +1,28 @@
+import { Request, Response } from 'express';
+import moment from 'moment';
+import { LiveCasesActions } from '../resources/actions/liveCaseActions';
+
+export default class LiveStatusController {
```
kurt-uk, line 5:
```suggestion
export default class LiveCaseStatusController {
```

.gitignore
```diff
@@ -25,3 +25,5 @@ dist/
 public/
 
 /coverage
+
+.env
```
kurt-uk, line 29: do we not need the env file youre adding?

src/main/controllers/LiveStatusController.ts
```diff
@@ -0,0 +1,28 @@
...
+  public get(req: Request, res: Response): void {
+    const courtId = req.query.courtId as string;
+    const timerMilliseconds = process.env.REFRESH_TIME_MILLISECONDS ? process.env.REFRESH_TIME_MILLISECONDS : 15000;
```
kurt-uk, line 8: default values can live in the declaration of the environment variable, that way we only need to change default in 1 place

src/main/routes/routes.ts
```diff
@@ -42,6 +42,8 @@ export default function(app: Application): void {
 
   app.get('/subscription-management', app.locals.container.cradle.subscriptionManagementController.get);
 
+  app.get('/live-status', app.locals.container.cradle.liveStatusController.get);
```
kurt-uk, line 45:
```suggestion
  app.get('/live-case-status', app.locals.container.cradle.liveStatusController.get);
```

src/main/views/live-status.njk
```diff
@@ -0,0 +1,84 @@
...
+                text: 'Status',
+                classes: 'table-header-bottom-align'
+            }
+        ],
+        rows: cases
+    }) }}
+    <hr class="govuk-header__container--full-width"/>
+    <a href="/view-option" class="govuk-link govuk-!-font-weight-bold">Find a court or tribunal list</a>
+    <script>
+      const timer = {{ refreshTimer }}
```
kurt-uk, line 79: is there any way we can add this kinda logic to the server side? if a user doesnt enable javascript then this wont work, which is fine for the resubmit stuff but this is the core functionality of the page

kurt-uk (changes requested):

ChrisS1512 (changes requested):
I've noticed the Find a court or tribunal list link doesn't work at the bottom.

It takes you to a view-option page, should this be search-option?

src/main/assets/scss/main.scss
```diff
@@ -123,3 +123,15 @@ tbody {
...
+
+.table-column-court-number {
+  width: 12%;
+}
+
+.table-header-bottom-align {
+  vertical-align: bottom;
+}
+
+.no-information-message {
```
ChrisS1512, line 135: No worries if not, however is there any way we can use the gov margin override classes for this instead

ChrisS1512 (changes requested):

src/main/controllers/LiveStatusController.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    if (courtId !== undefined) {
+      const liveCases = new LiveCasesActions().getLiveCases(parseInt(courtId));
+      if (liveCases) {
+        res.render('live-status', {
+          courtName: liveCases.courtName,
+          updateDateTime: moment.unix(liveCases.lastUpdated).format('MMMM Do YYYY, h:mm a'),
```
ChrisS1512, line 15: I've noticed the screenshot on the ticket has something like this:

Wednesday 30th July 2021       10:00am

where as the above would display as the following

Wednesday 30th July 2021, 10:00 am

Is the above change OK?

ChrisS1512 (changes requested):

.gitignore
commitBlob: It is recommended not to include .evn files into VCS as it might contain sensitive information.

src/main/controllers/LiveStatusController.ts
commitBlob: This is the only way I can set env variable on Windows machine

ChrisS1512 (approved):

src/main/controllers/LiveStatusController.ts
kurt-uk: we cant see what your variables are though? sensitive info isnt stored in the env file, wed have secrets placeholders so it would look something like:
```
REFRESH_TIME_MILLISECONDS=${REFRESH_TIME} | 15000
```

.gitignore
kurt-uk: we store all of the sensitive info in secrets

kurt-uk (commented):
you waiting till 514 is merged before you add the e2e journey in?

src/test/e2e/PageObjects/LiveCaseCourtSearchController.page.ts
```diff
@@ -1,12 +1,14 @@
+import { LiveCaseStatusPage } from './LiveCaseStatus.page';
+
 const helpers = require('../Helpers/Selectors');
 
 export class LiveCaseCourtSearchControllerPage {
   async getPageTitle(): Promise<string> {
-    $(helpers.pageTitle).catch(() => {
-      console.log(`${helpers.pageTitle} not found`);
+    $(helpers.CommonPageTitle).catch(() => {
```
kurt-uk, line 7: 👍🏼

src/test/e2e/tests/specs.test.ts
```diff
@@ -19,6 +20,9 @@ let searchPage: SearchPage;
 let searchResultsPage: SearchResultsPage;
 let subscriptionManagementPage: SubscriptionManagementPage;
 let liveCaseCourtSearchControllerPage: LiveCaseCourtSearchControllerPage;
+let liveCaseStatusPage: LiveCaseStatusPage;
+
+const validCourtName = 'Ailibugai Court';
```
kurt-uk, line 25: small thing, wont request changes for it but, can you move this into the describe for the flow its used in?

src/test/unit/views/live-case-status.test.ts
```diff
@@ -0,0 +1,75 @@
...
+  beforeAll(async () => {
+    await request(app).get(PAGE_URL).then(res => {
+      htmlRes = new DOMParser().parseFromString(res.text, 'text/html');
+    });
+  });
+
+  it('should display a back button with the correct value', () => {
+    const backLink = htmlRes.getElementsByClassName('govuk-back-link');
+    expect(backLink[0].innerHTML).contains('Back', 'Back button does not contain correct text');
+    expect(backLink[0].getAttribute('href')).equal('#', 'Back value does not contain correct link');
```
kurt-uk, line 21: this doesnt really apply if youre using history.back, so can we change the test to match?

kurt-uk (approved):

PR #84, opened 2021-09-22
PUB-740: Added subscription button in header

johnoneill46:
https://tools.hmcts.net/jira/browse/PUB-740

Added 'Subscription' button to header of all pages

### DIFF
--- a/src/test/unit/views/home.test.ts
+++ b/src/test/unit/views/home.test.ts
@@ -17,7 +17,9 @@ describe('Home page', () => {
 
   it('should display header',  () => {
     const header = htmlRes.getElementsByClassName(headingClass);
+    const nav = htmlRes.getElementsByClassName('govuk-header__navigation-item');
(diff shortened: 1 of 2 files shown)


src/test/unit/views/home.test.ts
```diff
@@ -17,7 +18,9 @@ describe('Home page', () => {
 
   it('should display header',  () => {
```
commitBlob, line 19: the header test you updated was actually checking for title header not the black bar header on top of the page (nav bar). Probably wording 'should display header' needs refactoring to 'should display title'
I would just add one more test rather  checking Subscriptions tag in the nav bar

johnoneill46: Absolutely right - changes made

kurt-uk (approved):

marcogagliardi (approved):
👌

PR #88, opened 2021-09-27
pub-744: my subscriptions

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-744

### Change description ###
- Added moj components to the project as some of them were required for this page (sub navigation - aka tabs, and sortable table)
- added mock data for users with subscriptions (userId 1) and without subscriptions (userId 2)
- refactored subscription management page and added tests

```
[ ] Yes
[x] No
```
all
case
court
none


src/main/views/subscription-management.njk
```diff
@@ -1,10 +1,115 @@
-{% from "./macros/common-components.njk" import goBack %}
+{% from "macros/common-components.njk" import goBack %}
```
ChrisS1512, line 1: When you first naviagate to the page, the case name column indicates that the data is sorted in ascending order, however on first load no sort is applied.

Clicking the sort arrow fixes this, however I think this should also be sorted on the initial load

.eslintignore
```diff
@@ -10,3 +10,5 @@ coverage/*
 /src/main/types/
 jest.*config.js
 *.min.js
+/src/main/assets/js/jquery.js
```
ChrisS1512, line 13: I believe JQuery should already be bundled as part of the package.json.

I think we should be able to remove this + the jquery file below

src/main/controllers/SubscriptionManagementController.ts
```diff
@@ -1,7 +1,34 @@
...
+import { SubscriptionActions } from '../resources/actions/subscriptionActions';
 
 export default class SubscriptionManagementController {
   public get(req: Request, res: Response): void {
-    res.render('subscription-management');
+    const subscriptionActions = new SubscriptionActions();
+    const caseTableData = subscriptionActions.generateCaseTableRows();
+    const courtTableData = subscriptionActions.generateCourtTableRows();
+    let activeAllTab, activeCaseTab, activeCourtTab;
+    switch (Object.keys(req.query)[0]) {
```
ChrisS1512, line 10: Only minor, however I think we can remove setting the 'false' values in each of the case statements, by setting the variables on line 9 initially to false.

src/main/views/subscription-management.njk
```diff
@@ -1,10 +1,115 @@
...
+                ],
+                rows: courtTableData
+            }) }}
+        {% endif %}
+    {% endif %}
+{% endblock %}
+
+{% block bodyEnd %}
+    <script type="text/javascript" src="/assets/js/jquery.js"></script>
+    <script type="text/javascript" src="/assets/js/mojAll.js"></script>
```
ChrisS1512, line 114: Is the mojAll needed here? I'm thinking it might already be included in the moj macro's

ChrisS1512 (changes requested):
The routes controller test will need updating with a few more tests to reflect the changes made in the controller

kurt-uk (changes requested):

src/main/views/subscription-management.njk
commitBlob: If I exclude these 2 files column sorting doesn't work

src/main/resources/mocks/userSubscriptions.json
```diff
@@ -0,0 +1,57 @@
+{
+  "results": [
```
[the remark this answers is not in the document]
commitBlob: I find it weird when you request json and get just array back

.eslintignore
commitBlob: I was not able to import already installed jQuery package into page.
I tried all importing suggestions from https://www.npmjs.com/package/jquery and none of them worked.

src/main/views/subscription-management.njk
commitBlob: I think sorting logic should be done on the BE, and we use mocked data. I will just change aria-sort to none for now? And once someone clicks on sort it should work as expected

.eslintignore
ChrisS1512: Ah ok. Are these two files the one's from the @ministryofjustice package? I'm wondering if we can copy and bundle these as part of webpack build rather than needing to store them as files in the repo separately. (We do this currently with images in the webpack config files)

src/main/views/subscription-management.njk
ChrisS1512: Sounds good :)

src/main/controllers/SubscriptionManagementController.ts
```diff
@@ -1,7 +1,27 @@
 import { Request, Response } from 'express';
+import { TableService } from '../service/tableService';
 
 export default class SubscriptionManagementController {
   public get(req: Request, res: Response): void {
-    res.render('subscription-management');
+    const userId = 1;
+    const tableService = new TableService();
```
kurt-uk, line 7: can this be class level declaration?

src/main/controllers/SubscriptionManagementController.ts
```diff
@@ -1,7 +1,27 @@
 import { Request, Response } from 'express';
+import { TableService } from '../service/tableService';
 
 export default class SubscriptionManagementController {
   public get(req: Request, res: Response): void {
-    res.render('subscription-management');
+    const userId = 1;
```
kurt-uk, line 6: i know this is only temporary but can this be out the method?

src/main/service/tableService.ts
```diff
@@ -0,0 +1,57 @@
+import moment from 'moment';
+import { SubscriptionActions } from '../resources/actions/subscriptionActions';
+
+export class TableService {
+  subscriptionActions = new SubscriptionActions();
+  
+  generateCaseTableRows(userId: number): any[] {
```
kurt-uk, line 7: can these methods take in data thats returned from api instead? so you dont have to do the call here, because then youre still limiting this method to only this situation, so something like: 
```
generateFourColumnTable(data)
```
then you can remove line 8 + 10 and this method is focused on creating a table with 4 columns

kurt-uk (changes requested):

kurt-uk (approved):

ChrisS1512 (approved):

src/main/service/tableService.ts
```diff
@@ -0,0 +1,51 @@
+import moment from 'moment';
+
+export class TableService {
```
kurt-uk, line 3: i was thinking this would be more generic, but if we cant, then lets change it back from TableService to SubscriptionSerivce, 
we cant use this service to create tables for anything else as is so maybe just rename

ChrisS1512 (approved):

PR #114, opened 2021-10-15
PUB-736: Build Subscriptions Case Name Search

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-736

### Change description ###
- add case name search and results page
- link subscription management page and add subscription pages 
- update e2e tests to follow add subscription management path
- fix invalid data mocks in hearings list

```
[ ] Yes
[x] No
```
search
search-error
search-results


src/main/views/case-name-search.njk
```diff
@@ -0,0 +1,81 @@
...
+    {% endif %}
+    {% set inputClasses = 'govuk-!-width-two-thirds govuk-input--error'
+        if noResultsError else 'govuk-!-width-two-thirds' %}
+    {% set formGroupError = 'govuk-form-group--error'
+        if noResultsError else 'govuk-form-group' %}
+    <form action="case-name-search" method="post">
+        <div class="govuk-form-group govuk-!-width-one-half">
+            <h1 class="govuk-heading-l">Enter a case name</h1>
+            {% if noResultsError %}
+                {{ govukInput({
```
marcogagliardi, line 44: Can we have only one input component? Instead of having a 2 input in a if then else block we can set some variables for the error message and error classes and use them to build the input component. See search.njk as an example

src/main/views/case-name-search.njk
```diff
@@ -0,0 +1,81 @@
...
+        text: text,
+        href: '#',
+        attributes: {
+            onClick: "history.back();event.preventDefault();"
+        }
+    }) }}
+{% endblock %}
+{% block content %}
+    {% if noResultsError %}
+        {% set errorMessage = 'Case name not found. Please try again or search by case reference number, unique reference number, or by a whole court' %}
```
marcogagliardi, line 21: Error message is different from the spec in the task description. Please see Ticket 6 section in the description.

src/main/views/case-name-search.njk
commitBlob: done

PR #117, opened 2021-10-19
Front end refactor

kurt-uk:

- [ ] commit messages are meaningful and follow good commit message guidelines
- [ ] README and other documentation has been updated / added (if needed)
- [ ] tests have been updated / new tests has been added (if needed)

## API
- removed the local api as this is not needed, calls will be made to the backend staging env by default, and env variables to be set for prod envs when needed
- service request files handle api calls and pass them to their appropriate service handler which then handles the bulk of the logic for the controller whilst the controller handles incoming and outgoing requests and responses and co-ordinates what part of the service to call
- no PIPApi to be passed across the frontend anymore, axios config is handled at the root level, not the top level 

## structure
- slight rename from Actions -> resources/requests which handle the api requests
- new services are to be configured in the axios config in the utils of the requests folder by creating instances of axios for required services

## Locales
content on njks have become placeholders where we can pass in the english locale in a json file. new files will need to be added to a PipRequest interface and json files naming should match this.

## mocks
mocks have been moved to testing

```
[ ] Yes
[ ] No
```


src/main/views/alphabetical-search.njk
```diff
@@ -3,7 +3,7 @@
 
 {% extends "template.njk" %}
 {% block pageTitle %}
-  {{ "Court List" | safe }}
+  {{ title }}
 {% endblock %}
 {% block beforeContent %}
   {{ goBack('/search-option') }}
```
commitBlob, line 9: Since this PR touches on many files we should probably change back button functionality to `onClick: "history.back();event.preventDefault();"` logic instead

src/test/e2e/tests/specs.test.ts
```diff
@@ -135,14 +132,13 @@ describe('Finding a court or tribunal listing', () => {
     });
 
     it('should display 3 result', async() => {
```
commitBlob, line 134: can we change should statement to 13 results as well

src/test/e2e/tests/specs.test.ts
```diff
@@ -135,14 +132,13 @@ describe('Finding a court or tribunal listing', () => {
...
 
   describe('Following the \'search\' path', () => {
-    const searchTerm = 'abergavenny';
-    const expectedNumOfResults = 1;
-    const expectedNumOfHearings = 1;
+    const searchTerm = 'Abergavenny Magistrates\' Court';
+    const expectedNumOfHearings = 13;
 
 
     it('should select \'tribunal hearing list\' option and navigate to search option page', async () => {
```
commitBlob, line 144: update should to reflect hearings list page instead of search option page since that step has been removed

commitBlob (changes requested):

src/test/e2e/tests/specs.test.ts
ChrisS1512: Updated :)

src/main/views/alphabetical-search.njk
ChrisS1512: Updated to use the new style of back button

ChrisS1512 (approved):

PR #118, opened 2021-10-20
pub-808 (Text changes)

junaidiqbalmoj:
Change the text and update the related tests

https://tools.hmcts.net/jira/browse/PUB-808

Image 20-10-2021 at 13 39

[ ] Yes
[x] No

### DIFF
--- a/src/main/views/search-option.njk
+++ b/src/main/views/search-option.njk
@@ -17,7 +17,7 @@
             name: "find-choice",
             fieldset: {
                 legend: {
-                    text: 'Do you know the name of the court or Tribunal?',
+                    text: 'Do you know the name of the court or tribunal?',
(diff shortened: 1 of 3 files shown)


src/main/views/search-option.njk
```diff
@@ -17,13 +17,13 @@
             name: "find-choice",
             fieldset: {
                 legend: {
-                    text: 'Find a court or tribunal list',
+                    text: 'Do you know the name of the court or Tribunal?',
```
ChrisS1512, line 20: Only minor however is it correct that Tribunal starts with a capital and court doesn't?

I'm thinking both should be lower case

ChrisS1512 (changes requested):

PR #147, opened 2021-11-05
PUB-859 date formats

furnivall:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-859

### Change description ###
Moved the time to lower line (in a slightly smaller font as it looked nicer - let me know if you want to switch it back).
Also altered unit tests to reflect new changes.
image

```
[ ] Yes
[ X] No
```


src/main/views/live-case-status.njk
```diff
@@ -16,7 +16,8 @@
 {% block content %}
     <h2 class="govuk-heading-l">Live hearing updates</h2>
```
kurt-uk, line 17: somehow looks like this line has been switched back, do you mind replacing with locale placeholder? it will look like 
```suggestion
    <h2 class="govuk-heading-l">{{ heading }}</h2>
```

src/main/views/live-case-status.njk
```diff
@@ -16,7 +16,8 @@
...
     <h3 class="govuk-heading-m">{{ courtName }}</h3>
-    <p class=" govuk-body-lead">{{ updateDateTime }}</p>
+    <p class=" govuk-body-lead">{{ updateDate }}</p>
+    <p class=" govuk-body">Last updated as at {{ updateTime }}</p>
```
kurt-uk, line 20: can you extract "Last updated as at " and replace with a nunjuks place holder? it looks the same as the above, and you would place it in `resources/locales/en/live-case-status.json` create a new attr. in there with this text

kurt-uk (changes requested):

src/main/views/live-case-status.njk
furnivall: Sorted.

src/main/views/live-case-status.njk
furnivall: Should also be done now

marcogagliardi (approved):
👌🏽

src/main/views/live-case-status.njk
```diff
@@ -14,9 +14,10 @@
 {% endblock %}
 
 {% block content %}
-    <h2 class="govuk-heading-l">Live hearing updates</h2>
+    <h2 class="govuk-heading-l">{{ heading }}</h2>
     <h3 class="govuk-heading-m">{{ courtName }}</h3>
-    <p class=" govuk-body-lead">{{ updateDateTime }}</p>
+    <p class=" govuk-body-lead">{{ updateDate }}</p>
+    <p class=" govuk-body">{{ lastUpdatedText }}{{ updateTime }}</p>
     <p class="govuk-body">The status of each hearing will be updated as events happen live in court</p>
```
commitBlob, line 21: Can you please use the locales value for this one as well

commitBlob (changes requested):

src/main/views/live-case-status.njk
furnivall: Hi Maro - should be done in just now's commit :)

kurt-uk (dismissed):

commitBlob (approved):

kurt-uk (dismissed):

commitBlob (approved):

ChrisS1512 (dismissed):

commitBlob (approved):

johnoneill46 (approved):

PR #158, opened 2021-11-09
Pub 876 mock session updates

furnivall:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-876

### Change description ###

Altered user types to "Verified Signed-in user" or "Unverified user (Not signed-in)"

Replaced all sign in buttons with "sign out" buttons for authenticated users within template.njk

Added "isAuthenticated" status to each controller with globalAuthGiver function within routes.ts

Added '/logout' route and logOut function to routes.ts (this functions destroys current session and redirects to '/' - please let me know if this is desired behaviour.

image
image

```
[ ] Yes
[X ] No
```


src/main/views/template.njk
```diff
@@ -9,6 +9,32 @@
 
 {% block header %}
 
+{% if isAuthenticated %}
```
commitBlob, line 12: Nice, I was just looking on 877 how to pass user context, I guess I can re-use your solution

src/main/views/template.njk
```diff
@@ -9,6 +9,32 @@
...
+{{ govukHeader({
```
commitBlob, line 13: Instead of rendering whole govukHeader in if else, can we just change last navigation item if user is signed in or not.

commitBlob (changes requested):

src/main/views/template.njk
furnivall: Absolutely - will sort now

johnoneill46 (approved):

commitBlob (approved):

PR #165, opened 2021-11-11
PUB-876 v2  - Removed unverified user radio option

furnivall:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-876

### Change description ###
After discussion with Nigel, we decided it made sense to remove the unverified user option entirely, as this status can be achieved with the "Sign out" button alone.

image

```
[ ] Yes
[ X] No
```


commitBlob:
Not sure what is the point of this radio in the end as there is only one user type 😂

commitBlob (approved):

johnoneill46 (approved):

PR #176, opened 2021-11-18
PUB-924: Accessibility timeout

commitBlob:
### Change description ###
As per Accessibility call increasing timeout to 30s 
https://github.com/pa11y/pa11y#timeout-number

```
[ ] Yes
[x] No
```

### DIFF
--- a/jest.a11y.config.js
+++ b/jest.a11y.config.js
@@ -7,6 +7,6 @@ module.exports = {
   },
   moduleFileExtensions: ['ts', 'js'],
   setupFiles: ["<rootDir>/jestEnvVars.js"],
-  // pa11y default
+  // pa11y recommended default
   testTimeout: 30000,
 }
--- a/jest.a11y.config.js
+++ b/jest.a11y.config.js
@@ -1,14 +1,12 @@
 module.exports = {
   roots: ['<rootDir>/src/test/a11y'],
-  "testRegex": "(/src/test/.*|\\.(test|spec))\\.(ts|js)$",
-   "moduleFileExtensions": [
-    "ts",
-    "js"
-  ],
-  "testEnvironment": "node",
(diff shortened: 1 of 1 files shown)


jest.a11y.config.js
```diff
@@ -1,14 +1,12 @@
...
+  testRegex: "(/src/test/.*|\\.(test|spec))\\.(ts|js)$",
+  testEnvironment: "node",
   transform: {
     '^.+\\.ts?$': 'ts-jest',
   },
-  moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
-  setupFiles: ["<rootDir>/jestEnvVars.js"]
+  moduleFileExtensions: ['ts', 'js'],
+  setupFiles: ["<rootDir>/jestEnvVars.js"],
+  // pa11y default
```
kurt-uk, line 10:
```suggestion
  // pa11y recommended default
```
if it was the pa11y default we wouldnt have to set it ;)

kurt-uk (dismissed):

kurt-uk (approved):

alberto-hortiguela (approved):
LGTM

kurt-uk:
sorry i thought that this was the recommended default timeout, not the actual default, surely if we dont specify then we should be hitting the 30 second anyway? if the default is 30 seconds and were hitting 5 seconds, then surely we have that coded somewhere? otherwise leaving it blank would mean that we have the 30 second default?

commitBlob:
In case anyone has same question as Kurt we discussed this on the call.
As we use Jest for testing it's default timeout is set to 5 seconds. This overrides pa11y default of 30 seconds, hence in the jest config we define Jest timeout for a11y tests to be 30 seconds.
https://github.com/pa11y/pa11y#timeout-number

PR #180, opened 2021-11-19
PUB-903 update infra to use new infra 

chris-pateman:
https://tools.hmcts.net/jira/browse/PUB-903

### Change description ###
Update to use the new infrastructure

```
[ ] Yes
[ X ] No
```


commitBlob (approved):
For a short term fix, I would agree to ignore this vulnerability as package-maintainers should fix it. 
In addition to above, I don't think passport-azure-ad is used currently as we use mock-session to log in.
We should probably monitor passport-azure-ad progress and future changes.

PR #182, opened 2021-11-23
PUB-936 workflow for github

kurt-uk:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-936

### Change description ###

added workflow yaml that will run yarn linting and will also run ci checks, Ci Checks can be removed if the benefit is not needed against the jenkins build

```
[ ] Yes
[x] No
```


.github/workflows/main.yml
```diff
@@ -0,0 +1,24 @@
...
+      - edited
+      - synchronize
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: deepakputhraya/action-pr-title@master
+      with:
+        regex: '^(PUB-[0-9]{3,4} [a-zA-Z0-9._\- ]+)|(\[Snyk\].+)$' # Regex the title should match.
```
alberto-hortiguela, line 16: I extended to 4 the digits just to ensure future PRs dont get blocked when we reach 4 digits
we can remove the 3digits if necessary in the future

PR #202, opened 2021-12-09
PUB-878 Added Redis credentials to frontend

adamtodd0:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-878

### Change description ###
Added the credentials for Redis cache.

**NOTE:**
As this is branched from master it includes tickets PUB-971 and PUB-980 which are already in master but not in pre-release

```
[ ] Yes
[x] No
```


README.md
```diff
@@ -69,11 +69,6 @@ Bundle:
 $ yarn webpack
 ```
 
-Start Redis:
```
chris-pateman, line 72: You could probably leave this in if they want to do local testing.
maybe add a sentence to say if you would like to use a local version then you can follow this guide for running a local redis or you can connect to the remote redis instance

src/main/cacheManager.ts
```diff
@@ -1,9 +1,21 @@
 import { Logger } from '@hmcts/nodejs-logging';
+import config from 'config';
+
+const redisPortalCredential = {
+  host: config.get('secrets.pip-ss-kv.REDIS_HOST'),
```
commitBlob, line 5: these will need to be defined in the config/default.json (can even be blanks), otherwise app crashes when run

src/main/cacheManager.ts
```diff
@@ -1,9 +1,21 @@
...
+  port: config.get('secrets.pip-ss-kv.REDIS_PORT'),
+  password: config.get('secrets.pip-ss-kv.REDIS_PASSWORD'),
+};
 
 const logger = Logger.getLogger('app');
+const connectionTest = redisPortalCredential;
+
+// const connection = process.env.NODE_ENV === 'production' ?
```
commitBlob, line 13: can we remove commented out code from the VCS please?

commitBlob (changes requested):

README.md
commitBlob: added command back to readme, resolving

commitBlob (approved):

commitBlob (dismissed):

commitBlob (approved):

alberto-hortiguela (approved):
LGTM

PR #218, opened 2022-01-07
PUB-979 - Add banner navigation unit tests

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-979

### Change description ###
New tests has been added covering both signed in and non-signed in cases. Tests are carried out on the view-option page which is part of the public journey.

No screenshots to provide as these are only additional unit tests

**NOTE:**
Some of the links point to the heroku app as they are used for the user testing.

```
[ ] Yes
[x] No
```


furnivall (approved):
👍

alberto-hortiguela (approved):
LGTM

PR #224, opened 2022-01-11
Pub-960 -  pi sign in page

marcogagliardi:
https://tools.hmcts.net/jira/browse/PUB-960

### Change description ###
pi authentication login and login return page placeholder
Screenshot 2022-01-11 at 13 11 08


alberto-hortiguela:
A couple of styling issues with the banner. 
1 There is some sort of left & top margin
2 It is missing the crown icon
image

alberto-hortiguela:
Can you also rename the flow to B2C_1_SignInUserFlow please?

src/main/resources/locales/en/login-return.json
```diff
@@ -0,0 +1,5 @@
+{
+  "title": "P&I Login",
+  "heading": "You have been logged in",
+  "div": "Your token is: "
```
kurt-uk, line 4: this isnt used??

kurt-uk (changes requested):

src/main/resources/locales/en/login-return.json
marcogagliardi: Yes true. I will delete it

marcogagliardi:
> A couple of styling issues with the banner. 1 There is some sort of left & top margin 2 It is missing the crown icon image

Good spot

src/test/unit/views/login-return.test.ts
```diff
@@ -0,0 +1,23 @@
...
+describe('Login return Page', () => {
+  beforeAll(async () => {
+    await request(app).get(PAGE_URL).then(res => {
+      htmlRes = new DOMParser().parseFromString(res.text, 'text/html');
+    });
+  });
+
+  it('should display header', () => {
+    const navigation = htmlRes.getElementsByClassName(navigationClass);
+    expect(navigation[0].innerHTML).contains(expectedHeader, 'Could not find the header');
```
kurt-uk, line 21: this test passes when it should fail now as youre missing the header which means its an empty page, so the test isnt actually testing what you think it is, it will need refactoring to make sure that it fails when the header isnt displayed and only passes when the header is displayed

kurt-uk: just in case its not clear:
this is what the page is rendering:
Screenshot 2022-01-17 at 15 20 33

and this is your test passing when it should fail:

Screenshot 2022-01-17 at 15 21 23

kurt-uk (changes requested):

src/main/resources/locales/en/login-return.json
```diff
@@ -0,0 +1,3 @@
+{
+  "title": "P&I Login"
```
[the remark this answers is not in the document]
marcogagliardi: I have added the heading back. Bear in mind this is just a placeholder as the returning page needs to be developed. At the moment Nigel needs a discussion for it to see what they want.

src/test/unit/views/login-return.test.ts
marcogagliardi: added the heading and added the test

kurt-uk (approved):

alberto-hortiguela (approved):
LGTM

alberto-hortiguela (approved):
LGTM

commitBlob:
Don't think we need controller for login return page as I was able to handle authentication without it see PR #235 
Therefore, believe this PR can be closed

alberto-hortiguela:
No need for this PR.
Working under https://tools.hmcts.net/jira/browse/PUB-1028

PR #258, opened 2022-02-07
PUB-1035 - Account Home Amends

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-1035

### Change description ###
- Display 3 cards per row
- remove LCSU card
- amend text on the cards

```
[ ] Yes
[x] No
```
accounthome


src/main/assets/scss/main.scss
```diff
@@ -358,8 +358,8 @@ div {
   flex-wrap: wrap;
 
   > div {
-    flex: 50%;
-    max-width: 50%;
+    flex: 33.33%;
```
kurt-uk, line 361: would the gov width overrides not work here?

kurt-uk (changes requested):

src/main/assets/scss/main.scss
commitBlob: yes and no.
I can use `govuk-!-width-one-third` and that will display 3 in a row, however doesn't play well with loops, as if there are more elements they're just added in the row, they don't break down.

marcogagliardi (approved):
👌🏽

kurt-uk (approved):

kurt-uk (approved):

alberto-hortiguela (approved):

PR #262, opened 2022-02-09
PUB-1062 - AAD on e2e

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-1062

### Change description ###
- use AAD process in e2e
- refactor user id to oid as id doesn't exist on AAD

```
[ ] Yes
[x] No
```


alberto-hortiguela (dismissed):

charts/pip-frontend/values.stg.template.yaml
```diff
@@ -3,3 +3,4 @@ nodejs:
   image: ${IMAGE_NAME}
   ingressHost: ${SERVICE_FQDN}
   environment:
+    EXCLUDE_E2E: true
```
kurt-uk, line 6: nooo???

src/main/controllers/PendingSubscriptionsController.ts
```diff
@@ -7,9 +7,10 @@ const subscriptionService = new SubscriptionService();
 
 export default class PendingSubscriptionsController {
   public async get(req: PipRequest, res: Response): Promise<void> {
+    console.log('get', req.user['oid']);
```
kurt-uk, line 10: left console log in

ChrisS1512 (approved):

kurt-uk (changes requested):

src/main/controllers/PendingSubscriptionsController.ts
alberto-hortiguela: this could be removed

charts/pip-frontend/values.stg.template.yaml
alberto-hortiguela: we need to exclude by now, tech debt ticket was created

alberto-hortiguela (approved):

charts/pip-frontend/values.stg.template.yaml
alberto-hortiguela: my wrong, staging should have it

kurt-uk (approved):

src/main/controllers/PendingSubscriptionsController.ts
commitBlob: I noticed it as well and removed it along enabling e2e on staging

PR #264, opened 2022-02-10
PUB-1072 - File Too Large Handler

commitBlob:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-1072

### Change description ###
Create a middleware which will catch multer error for files which exceed threshold limit.
All other multer errors will render error page instead of displaying generic error.
Changes on files: 
- routes.ts fileSizeLimitErrorHandler middleware
- manualUploadController show error if else block

```
[ ] Yes
[x] No
```


ChrisS1512 (changes requested):
Looks good! 

I think we should add a test for the file upload, just to check the error message is displayed when it's above 2MB

ChrisS1512 (dismissed):

kurt-uk (dismissed):

PR #309, opened 2022-03-11
ON HOLD - PUB-1054 - screen reader handling empty cells

marcogagliardi:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-1054

### Change description ###
Added aria decorators


commitBlob (changes requested):
We already tried aria hidden an it makes the issue even worse. If there are 2 courts starting with A it will read row A  name of the court, row B with name of the court even though the name of the second court still starts with A

Captures

furnivall (approved):
Nice one dude :)

marcogagliardi:
@commitBlob are we talking about empty cells in a table? Because the issue it is that screen readers don't like empty headers cells. this error only shows when there's no non-empty td associated with the th. I changed the empty th cell in td. I have tested with the screen reader and it looks like it is working.

commitBlob:
> @commitBlob are we talking about empty cells in a table? Because the issue it is that screen readers don't like empty headers cells. this error only shows when there's no non-empty td associated with the th. I changed the empty th cell in td. I have tested with the screen reader and it looks like it is working.

Have you read the comment on the ticket?
I'll let you be it your way and QA fail you.

commitBlob (approved):

ChrisS1512:
Closing this PR as covered under 1205

PR #312, opened 2022-03-15
PUB-1166 added BtT to template and extended all components need from it

kurt-uk:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-1166

### Change description ###
added BtT to template and extended all components need from it and fixed cookie naming 
BtT appears on most pages, the look is the same
Screenshot 2022-03-15 at 14 46 51
Screenshot 2022-03-15 at 14 47 03

```
[ ] Yes
[x] No
```


furnivall (dismissed):

alberto-hortiguela (dismissed):

commitBlob (dismissed):
mkovb

marcogagliardi (changes requested):
Can we use the parent-box wrapper in the template so we don't need to have it in each page?

kurt-uk:
> Can we use the parent-box wrapper in the template so we don't need to have it in each page?

unfortunately not, that would only work on about 15% of our pages, as each page is done slightly differently in how its holds context of its content, some of the pages such as the alphabetical search page would work, where the content isnt isolated. The rest of our pages though contain their content in a way where if we exclude the parent box from the content (as putting the div in the template would mean as you cant then inject the content mid way in the template) the 2 are treated as separate entities and doesnt display properly

alberto-hortiguela:
Conflict issues

alberto-hortiguela (dismissed):

commitBlob (approved):

ChrisS1512 (approved):
Looks good. Just needs updating with base branch

alberto-hortiguela (dismissed):

ChrisS1512 (dismissed):

ChrisS1512 (approved):

joshblackmoor (approved):

ChrisS1512 (approved):

furnivall (approved):

PR #389, opened 2022-05-16
Pub 1248/1220 update Subs summary screen and add another subscription link

joshblackmoor:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1248
https://tools.hmcts.net/jira/browse/PUB-1220

### Change description ###

Changes I have made for this ticket:

- Updated ‘your subscriptions’ to ‘Your email subscriptions’
- Updated green button from ‘Add new subscription’ to ‘Add email subscription’ 
- Updated ‘You currently have no subscriptions’ to ‘You do not have any active subscriptions’
- Added logic to hide the filters when there are zero subscriptions (only ‘You do not have any active subscriptions’ text is shown)
- Removed sorting on the headers

Have also done PUB-1220 as part of this ticket. 
Changes:

- Renamed the link from ‘Add another’ to ‘Add another email subscription’
- Moved the link above the ‘Confirm Subscriptions’ button
- Aligned the text size with the ‘Remove’ link on that page

```
[ ] Yes
[x] No
```


junaidiqbalmoj (commented):
I think you need to make changes in e2e tests as well.

src/main/views/pending-subscriptions.njk
```diff
@@ -86,9 +86,10 @@
         <p class="govuk-body govuk-!-margin-bottom-8">No pending court subscriptions</p>
       {% endif %}
 
+
       <div class="govuk-!-text-align-left">
+        <a class="govuk-link" href="{{ addOther_link }}">{{ addOther_text }}</a>
```
ChrisS1512, line 91: I think the text should be above the button (rather than to the left of it) for 1220. Would you be able to double check?

ChrisS1512 (changes requested):

src/main/views/pending-subscriptions.njk
joshblackmoor: Done :)

ChrisS1512: Hmm still seems to be the same on mine.

image

Would you be able to have another look?

joshblackmoor: Screenshot 2022-05-16 at 10 47 48

Wrapped it in a P tag to give it the spacing it needs and added govuk-body so it matches the 19px size of the "Remove" link above.

kurt-uk (approved):

ChrisS1512 (approved):

PR #411, opened 2022-05-27
PUB-1345 - Fix dot filenames

furnivall:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-1345

### Change description ###
Pre-fix:
The following error is thrown when any file with a dot before the file extension is uploaded. 

image

image

This also applies to the manual upload service: 

image

Problem is now fixed, hopefully.

```
[ x] Yes
[ ] No
```


src/main/service/fileHandlingService.ts
```diff
@@ -43,7 +43,7 @@ export class FileHandlingService {
   }
 
   isValidFileType(fileName: string, image: boolean): boolean {
-    const fileType = fileName.split('.')[1]?.toLocaleLowerCase();
+    const fileType = fileName.slice((fileName.lastIndexOf('.') - 1 >>> 0) + 2).toLocaleLowerCase();
```
kurt-uk, line 46: would fileName.slice((fileName.lastIndexOf('.')) +1) not work and be less conversion? rather than converting to into which index already returns?

kurt-uk (changes requested):

src/main/service/fileHandlingService.ts
furnivall: So - I ran a couple of tests on jsfiddle to check this. The complexity is there to handle the edge case where a file starts with a dot and has no extension (e.g. .gitignore or .htaccess). In those specific cases, the bit shift operator hack allows us to prevent it from returning anything (e.g. if you were to submit a file called ".png" or something to that effect, it would return an invalid filetype, whereas with the +1, it would allow that to be submitted).

It all comes down to what you're looking for, really.

kurt-uk (approved):

joshblackmoor (approved):

PR #425, opened 2022-06-10
PUB 1290 cookies page updates

joshblackmoor:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1290

### Change description ###

- Updated cookies page with new wording
- Added welsh version of the cookie page
- Added form to enable/disable some cookies and set up analytics and performance cookie functions to delete those cookies

```
[ ] Yes
[x] No
```


src/main/resources/locales/cy/cookie-policy.json
```diff
@@ -0,0 +1,86 @@
+{
```
ChrisS1512, line 1: Not specific to this file, however alongside the below we should also ensure that the cookies are disabled when the user declines them.

I believe theres code elseware in this repo that we can plugin to do this.

ChrisS1512 (changes requested):

kurt-uk:
cant see where the logic is that uses the functionality of actually disabling or enabling depending on the toggles; see cookie-preferences.js

ChrisS1512 (changes requested):
Currently getting a sonarqube failure on the build. Once sorted I'll give this a test

ChrisS1512:
Just trying this out now via the DEV pod and seems to be keeping the cookies. Would you be able to try it out to see if you get the same result?

furnivall (dismissed):
Really great work guys - <3

ChrisS1512 (dismissed):

ChrisS1512 (approved):

junaidiqbalmoj (approved):

PR #428, opened 2022-06-15
PUB-1303 Header changes to Email Subs page

junaidiqbalmoj:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1303

### Change description ###

Changes to "Confirm your email subscriptions" and "Your email subscriptions".

Image 15-06-2022 at 16 51

Image 15-06-2022 at 16 50

[] Yes
[x] No


src/main/views/subscription-management.njk
```diff
@@ -120,4 +120,18 @@
   {{ super() }}
   <script type="text/javascript" src="/assets/js/jquery.js"></script>
   <script type="text/javascript" src="/assets/js/mojAll.js"></script>
+
+  <script>
+    window.onload = function() {
```
ChrisS1512, line 125: Is there a way we could do the below using Nunjucks instead of JS? Where possible I think it's better to render it via the server side instead of the frontend JS

src/main/views/pending-subscriptions.njk
```diff
@@ -40,48 +40,52 @@
     </form>
```
ChrisS1512, line 40: Needs some unit tests adding for this + the other file

ChrisS1512 (changes requested):

src/main/views/pending-subscriptions.njk
junaidiqbalmoj: unit tests have been added for both pending and confirmation subscription pages.

kurt-uk (dismissed):

ChrisS1512 (approved):

kurt-uk (approved):

PR #439, opened 2022-06-28
PUB-1419 PartyRole mappings

joshblackmoor:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1419

### Change description ###

Added mapping for party roles.

```
[ ] Yes
[x] No
```


src/main/service/publicationService.ts
```diff
@@ -252,4 +254,16 @@ export class PublicationService {
     }
     return publishedTime;
   }
+
+  private static convertPartyRole(nonConvertedPartyRole: string): string {
```
ChrisS1512, line 258: Need some unit tests for this bit :)

ChrisS1512 (changes requested):

src/main/service/publicationService.ts
joshblackmoor: 👍 Have added a test to cover a list that has values to be mapped (sonar is now showing 100% coverage for this PR). 
I've also refactored the function again to only use one if as realised it was unnecessary for there to be two separate ones.

ChrisS1512: Really nice :)

ChrisS1512 (approved):

furnivall (approved):
Great code, I love it. Such sexy code.

ChrisS1512 (approved):

furnivall (approved):

PR #445, opened 2022-07-04
PUB-1428 Unshuttering Changes

furnivall:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-1428

### Change description ###
Some simple frontend changes for unshuttering.

- Temporarily altered the for loop on home/interstitial page to only show the second bullet
- Fixed the home banner item to go to home rather than view-option
- Temporarily altered the example text on the search page to show sjp instead of oxford combined court.

I've preserved code where possible to make it easy to revert according to business need.

```
[ ] Yes
[x ] No
```


joshblackmoor (approved):
LGTM 👍

src/main/views/template.njk
```diff
@@ -89,7 +89,7 @@
     [
       {
         text: navLinks.link1,
-        href: '/view-option'
+        href: '/'
```
ChrisS1512, line 92: As discussed I reckon this will also need doing on the sign in pages

ChrisS1512 (changes requested):

src/main/views/template.njk
furnivall: done now :)

ChrisS1512 (approved):

PR #447, opened 2022-07-06
PUB-1429 All party details fixes civil and family lists

junaidiqbalmoj:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1429

### Change description ###

Fix the bugs which show undefined on the page if certain information is not provided.

Image 06-07-2022 at 13 07

[ ] Yes
[x] No

### DIFF
--- a/src/main/service/publicationService.ts
+++ b/src/main/service/publicationService.ts
@@ -122,8 +122,8 @@ export class PublicationService {
   }
 
   private findAndManipulatePartyInformation(hearing: any): void {
-    let applicant;
-    let respondent;
+    let applicant = '';
+    let respondent = '';
 
     if(hearing?.party) {
       hearing.party.forEach(party => {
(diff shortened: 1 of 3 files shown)


src/main/service/publicationService.ts
```diff
@@ -122,44 +122,44 @@ export class PublicationService {
...
             applicant += this.stringDelimiter(applicant?.length, ',');
             break;
           }
           case 'APPLICANT_PETITIONER_REPRESENTATIVE':
           {
-            applicant += party?.friendlyRoleName;
-            applicant += this.stringDelimiter(party?.friendlyRoleName.length, ':');
-            applicant += this.createIndividualDetails(party.individualDetails);
+            applicant += this.writeStringIfValid(party?.friendlyRoleName);
+            applicant += this.stringDelimiter(this.writeStringIfValid(party?.friendlyRoleName).length, ':');
```
joshblackmoor, line 141: Instead of friendlyRoleName it should just be  'Legal Advisor:' now

src/main/service/publicationService.ts
```diff
@@ -122,44 +122,44 @@ export class PublicationService {
...
             respondent += this.stringDelimiter(respondent?.length, ',');
             break;
           }
           case 'RESPONDENT_REPRESENTATIVE':
           {
-            respondent += party?.friendlyRoleName;
-            respondent += this.stringDelimiter(party?.friendlyRoleName.length, ':');
-            respondent += this.createIndividualDetails(party.individualDetails);
+            respondent += this.writeStringIfValid(party?.friendlyRoleName);
+            respondent += this.stringDelimiter(this.writeStringIfValid(party?.friendlyRoleName).length, ':');
```
joshblackmoor, line 154: Same as above re friendlyRoleName

joshblackmoor (changes requested):

src/main/service/publicationService.ts
junaidiqbalmoj: Yes, you are right. I hardcode the LEGALADVISOR in code, but it will be replaced with English or Welsh in nunjucks. That's how we are doing for some other text as well.

joshblackmoor (approved):
Tried it out locally, good job 👍

ChrisS1512 (approved):

joshblackmoor (approved):

PR #448, opened 2022-07-06
PUB-1425 - Role Differentiation

ChrisS1512:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1425

### Change description ###

This PR provides role differentiation between the admin users.

It also locks down the subscriptions flows so that only media users can get access to them

```
[ ] Yes
[x] No
```


joshblackmoor (approved):
Really nice code 🥇

src/main/views/admin-dashboard.njk
```diff
@@ -13,19 +13,58 @@
...
+            <a href="{{ uploadCard.link }}"
+               class="govuk-link govuk-body govuk-!-font-weight-bold"
+               id="card-{{ uploadCard.link }}"
+            >{{ uploadCard.title }}</a>
+          </p>
+          <p class="govuk-body">{{ uploadCard.description }}</p>
+        </div>
+      </div>
+      {% endif %}
+      {% if user.isAdmin %}
```
KianKwa, line 29: It is very minor....guess we don't need a further 'if' condition here as it can be combined with the previous "if"?

KianKwa (approved):

src/main/views/admin-dashboard.njk
ChrisS1512: Good spot :) Now combined with the previous if statement

KianKwa (dismissed):

joshblackmoor (dismissed):

KianKwa (approved):

furnivall (approved):

PR #450, opened 2022-07-07
PUB-1392 - Reset Password Custom Policy

furnivall:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-1392

### Change description ###
Implemented custom policy in azure to replace the built-in reset password flow with a new version that checks user status before being allowed to send reset emails and implemented logic to handle in frontend.

This change essentially involves handling the AADB2C90118 error within AD B2C on our application and then redirecting to the newly built fix.

```
[ ] Yes
[x] No
```


src/main/authentication/adminAuthentication.ts
```diff
@@ -13,4 +18,19 @@ export class AdminAuthentication {
     }
     return false;
   }
+
+  public forgotPasswordRedirect(req, res, next): void {
+    const body = JSON.stringify(req.body);
+    if (body.includes('AADB2C90118')) {
+      const CLIENT_ID = config.get('secrets.pip-ss-kv.CLIENT_ID');
```
KianKwa, line 25: Hey Danny, I think it looks a bit easier to read if all the URL query parameters are laid out in json. It is not crucial to change but just a suggestion. It is up to u :) 

const params = new URLSearchParams({
  ‘p’: `${authenticationConfig.FORGOT_PASSWORD_POLICY}`,
  'client_id': config.get('secrets.pip-ss-kv.CLIENT_ID'),
  ‘nonce’: ‘defaultNonce’,
  'redirect_uri': `${FRONTEND_URL}/password-change-confirmation`,
  'scope': ‘openid’,
  ‘responseType’:’id_token’,
  ‘prompt’: 'login'
});
res.redirect(`${B2C_URL}/oauth2/v2.0/authorize? ` + params.toString());

furnivall: Hey Kian - I 100% agree with you on this, it's a much nicer solution. However, I'm concerned about the limited time we have, so I will add it in my next (less stressfully timed) frontend PR - good compromise?

KianKwa: Sure thing. Will resolve the comment :)

joshblackmoor (approved):

PR #453, opened 2022-07-07
PUB 1276 - Welsh translation for court page

junaidiqbalmoj:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1276

### Change description ###

Make courts A-Z page working for Welsh language

Image 07-07-2022 at 17 58

[ ] Yes
[x] No


furnivall (approved):

src/main/service/manualUploadService.ts
```diff
@@ -52,27 +52,37 @@ export class ManualUploadService {
...
+  private async validateCourt(courtName: string, language: string): Promise<string> {
     if (courtName?.length >= 3) {
-      const validCourt = await courtService.getLocationByName(courtName);
+      const validCourt = await courtService.getLocationByName(courtName, language);
       if (validCourt) {
         return null;
       }
-      return 'Please enter and select a valid court';
+
+      if (language == 'cy') {
```
KianKwa, line 75: It might be a lot of effort to change it, or might not be possible. Just thinking out loud....can those error strings not going into the language resource files and we use the appropriate ones based on req.lng. This should put all translations in one place rather than hardcoding them?

KianKwa (dismissed):

KianKwa (approved):

PR #455, opened 2022-07-08
PUB-1449 Fix bug with Duration on all daily lists

junaidiqbalmoj:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1449

### Change description ###

Calculate the correct duration for all the daily lists

Image 08-07-2022 at 11 46

[ ] Yes
[x] No


joshblackmoor (approved):

src/main/service/publicationService.ts
```diff
@@ -76,7 +76,7 @@ export class PublicationService {
       let durationAsHours = 0;
       let durationAsMinutes = moment.duration(sittingEnd.startOf('minutes').diff(sittingStart.startOf('minutes'))).asMinutes();
       if (durationAsMinutes >= 60) {
-        durationAsHours = moment.duration(sittingEnd.startOf('hours').diff(sittingStart.startOf('hours'))).asHours();
+        durationAsHours = Math.floor(durationAsMinutes / 60);
```
KianKwa, line 79: Junaid, is that the intention to change it to the smaller number? If we have 1 hour and 59 minutes this will be rounded down to 1 hour. If that is by design that is fine. If not, may be Mach.round() could be used instead?

PR #496, opened 2022-08-10
PUB-1474 Welsh translation

junaidiqbalmoj:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1474

### Change description ###

Welsh translation for create media user error messages and other pages.

Image 10-08-2022 at 11 30

[ ] Yes
[x] No


joshblackmoor (approved):

src/main/service/createAccountService.ts
```diff
@@ -32,48 +34,53 @@ const accountManagementRequests = new AccountManagementRequests();
 const fileHandlingService = new FileHandlingService();
 
 export class CreateAccountService {
-  public validateFormFields(formValues: object, file: File): object {
+  public validateFormFields(formValues: object, file: File,
+    language: string, languageFile: string): object {
     return {
       nameError: {
-        message: this.validateMediaFullName(formValues['fullName']),
+        message: this.validateMediaFullName(formValues['fullName'], language, languageFile),
```
KianKwa, line 41: When getting the text for the error within each of the validation method, you are calling the language parser getText() method. This will read from the file every time. I think it might be better to read the json from the file once at the beginning of this method, pass down the json into each validation method, and create another method for getText() in the language parser which does not involves reading from file every time.

src/main/service/manualUploadService.ts
```diff
@@ -54,37 +57,30 @@ export class ManualUploadService {
     return [{text: 'SJP Media Register', value: 'SJP_MEDIA_REGISTER'}];
   }
 
-  public async validateFormFields(formValues: object, language: string): Promise<object> {
+  public async validateFormFields(formValues: object, language: string, languageFile: string): Promise<object> {
     const fields = {
-      courtError: await this.validateCourt(formValues['input-autocomplete'], language),
-      contentDateError: this.validateDate(this.buildDate(formValues,'content-date-from'), language),
+      courtError: await this.validateCourt(formValues['input-autocomplete'], language, languageFile),
```
KianKwa, line 62: This has the same issue with the previous comment in CreateAccountService. It is reading the json from file 3 times. We can just read the json once before getting the text from the json

src/main/service/createAccountService.ts
junaidiqbalmoj: agreed... and fixed the code

src/main/service/manualUploadService.ts
junaidiqbalmoj: agreed... and fixed the code

KianKwa (dismissed):

joshblackmoor (dismissed):

ChrisS1512 (approved):

joshblackmoor (approved):
👍

PR #497, opened 2022-08-11
PUB 1318 Media verification flow

joshblackmoor:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1318

### Change description ###

- Added new media verification route used for b2c login
- Created new return endpoint to ensure the user came through the verification flow (if a user had already signed in with the normal user flow there is a chance the returned flow in the user data is incorrect, for 60 mins. This is why a new return was created)
- Added new passport strategy for the flow
- Added new banner that shows on the account home page after an account has been verified
- Added new request to account management to update a media users last verified date when they come through the new flow

Screenshot 2022-08-11 at 10 37 48

```
[ ] Yes
[x] No
```


KianKwa (approved):

KianKwa (approved):

charts/pip-frontend/values.yaml
```diff
@@ -7,6 +7,7 @@ nodejs:
   environment:
     OIDC: true
     AUTH_RETURN_URL: https://127.0.0.1/login/return
+    MEDIA_VERIFICATION_RETURN_URL: https://127.0.0.1/media-verification/return
```
ChrisS1512, line 10: This will fail in prod non dev or staging environments, as it is not overridden. Needs putting in sds-flux-config :)

ChrisS1512 (changes requested):

charts/pip-frontend/values.yaml
joshblackmoor: Good point. Have added them, see: https://github.com/hmcts/sds-flux-config/pull/2173

ChrisS1512 (dismissed):

ChrisS1512 (approved):

KianKwa (approved):

PR #499, opened 2022-08-12
PUB-1518 Temp remove create account link

joshblackmoor:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1518

### Change description ###

Commented out the create account link and tests temporarily until we are ready to add it back in.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/views/sign-in.njk
+++ b/src/main/views/sign-in.njk
@@ -54,12 +54,14 @@
         ]
       }) }}
         {{ submitButton(continueButton) }}
-      <div class="govuk-grid-row govuk-!-margin-bottom-2">
-        <div class="no_padding govuk-grid-column-full">
-          <p class="govuk-body-s govuk-!-font-weight-bold govuk-!-margin-bottom-1">{{ accountInfo }}</p>
-          <a class="govuk-link" href="{{ accountInfo_link }}">{{ accountInfo_link_text }}</a>
-        </div>
-      </div>
+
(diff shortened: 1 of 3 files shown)


src/main/views/sign-in.njk
```diff
@@ -54,12 +54,14 @@
         ]
       }) }}
         {{ submitButton(continueButton) }}
-      <div class="govuk-grid-row govuk-!-margin-bottom-2">
-        <div class="no_padding govuk-grid-column-full">
-          <p class="govuk-body-s govuk-!-font-weight-bold govuk-!-margin-bottom-1">{{ accountInfo }}</p>
-          <a class="govuk-link" href="{{ accountInfo_link }}">{{ accountInfo_link_text }}</a>
-        </div>
-      </div>
+
```
ChrisS1512, line 57: Can I double check we're OK leaving the route accessible too? Or should the route also be disabled

ChrisS1512 (changes requested):

src/main/views/sign-in.njk
joshblackmoor: Checked with NC and leaving route accessible is fine, only want it temp removed off page

ChrisS1512: Ok great :)

ChrisS1512 (approved):

KianKwa (approved):

PR #510, opened 2022-08-23
PUB-1512 Confirmation screen for list type

junaidiqbalmoj:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1512

### Change description ###

Add confirmation screen for list type subscription.

Image 25-08-2022 at 16 03

[ ] Yes
[x] No


src/main/service/subscriptionService.ts
```diff
@@ -199,7 +204,151 @@ export class SubscriptionService {
     return payload;
   }
 
+  public async configureListTypeForLocationSubscriptions(userId, listType): Promise<boolean> {
+    return await subscriptionRequests.configureListTypeForLocationSubscriptions(
+      this.createListTypeSubscriptionPayload(userId, listType));
+  }
+
+  createListTypeSubscriptionPayload(userId, listType): object {
```
ChrisS1512, line 212: This method can be marked as private

src/main/service/subscriptionService.ts
```diff
@@ -199,7 +204,151 @@ export class SubscriptionService {
...
+    let listTypeArray;
+    if(listType) {
+      if (!Array.isArray(listType)) {
```
ChrisS1512, line 215: Can this happen? (should it always be either an array, or blank)

ChrisS1512 (changes requested):

src/main/service/subscriptionService.ts
junaidiqbalmoj: I implemented it in the same way function createSubscriptionPayload already there, but I changed it to private now :)

src/main/service/subscriptionService.ts
junaidiqbalmoj: I implemented it in the same way function createSubscriptionPayload already there, but I changed it to private now :)

ChrisS1512 (approved):

src/main/resources/locales/en/subscription-configure-list-confirmed.json
```diff
@@ -0,0 +1,12 @@
+{
+  "title": "Subscriptions updated",
+  "header": "Subscriptions updated",
+  "subheader": "Your subscription(s) have been updated successfully",
```
KianKwa, line 4: Can you double check the grammar here. The ticket says "has been" but it is "have been" here :)

src/main/resources/locales/en/subscription-configure-list-confirmed.json
```diff
@@ -0,0 +1,12 @@
...
+  "p1": "To continue, you can go to ",
+  "link_text": "your account",
+  "p2": " in order to:",
+  "addEmailSubscription": "add a new email subscription",
+  "manageYourCurrentSubscriptions": "manage your current email subscriptions",
+  "manageYourListTypeSubscriptions": "Further refine the list types per court received",
```
KianKwa, line 10: This is part of the bullet points like the other lines but this starts with an upper case but the other ones start with lower case. Consider making them consistent?

joshblackmoor, line 10: Just wondering if this should be a capital or not?

The mock up does show it but why are the rest on the page lower case.

Seems like we either need all with capital or none?

KianKwa (approved):

src/main/resources/locales/en/subscription-configure-list-confirmed.json
junaidiqbalmoj: As per discussion with Chris C, we do not need to include third bullet point. So, I have removed and and updated the rest of content changes on the page.

junaidiqbalmoj: As per discussion with Chris C, we do not need to include third bullet point. So, I have removed and and updated the rest of content changes on the page.

joshblackmoor (approved):
Looks good!

Just the Welsh that is missing from the changes you've made which I assume will be covered in a future ticket.

furnivall (dismissed):

joshblackmoor (approved):

KianKwa (approved):

PR #518, opened 2022-08-30
PUB-1526 Add SJP press register

KianKwa:
### JIRA link ###

Add SJP press register

### Change description ###

https://tools.hmcts.net/jira/browse/PUB-1526

```
[ ] Yes
[x] No
```


sonar-project.properties
```diff
@@ -6,5 +6,5 @@ sonar.sources=src/main
 sonar.tests=src/test
 sonar.test.exclusions=test/coverage/**
 sonar.javascript.lcov.reportPaths=coverage/lcov.info
-sonar.exclusions=src/main/app.ts,src/main/development.ts,src/main/index.js,src/main/server.ts,src/main/modules/i18next/index.ts,src/main/bundles/cookie-preferences.ts,src/main/modules/helmet/index.ts
+sonar.exclusions=src/main/app.ts,src/main/development.ts,src/main/index.js,src/main/server.ts,src/main/bundles/cookie-preferences.ts,src/main/modules/**/*.ts
```
ChrisS1512, line 9: I'd avoid the wildcard exclusion here. Having them defined individually means any modules added in the future will correctly fail unless explicitly excluded

ChrisS1512 (changes requested):

sonar-project.properties
furnivall: This will also be excluded in my PR for ... 1509? I think

PR #533, opened 2022-09-26
PUB-1556 Add missing Welsh translations

KianKwa:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1556

### Change description ###

Add missing Welsh translations

```
[ ] Yes
[x] No
```


ChrisS1512 (dismissed):

joshblackmoor (approved):

src/main/resources/locales/cy/common.json
```diff
@@ -0,0 +1,14 @@
+{
```
furnivall, line 1: This feels like it should be a custom nunjucks filter (look in src/modules/nunjucks/index.ts for some examples). This way you could do something like this as a filter

env.addFilter('month', function(x, y){x == 'en' ? y : monthLookup[y];});

and then to call in frontend, it'd be like 

{{ month | month(lng) }}

furnivall (dismissed):

ChrisS1512 (approved):

joshblackmoor (approved):

PR #538, opened 2022-09-30
PUB-1540 Updated subscription configure list screen with extra information

KianKwa:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1540

### Change description ###

Updated subscription configure list screen with extra information, including some Welsh translations.

```
[ ] Yes
[x] No
```


src/main/resources/welshJurisdictionLookup.json
```diff
@@ -0,0 +1,7 @@
+{
+  "Civil": "Llys Sifil",
```
ChrisS1512, line 2: I'm wondering if this should be part of listLookup rather than a seperate file.

It does mean a bit of duplication, bit feels a bit cleaner as they're so closely related

src/main/views/subscription-configure-list.njk
```diff
@@ -14,6 +14,8 @@
 {% block content %}
     <h1 class="govuk-heading-l">{{ title }}</h1>
```
ChrisS1512, line 15: I'm getting an issue in the view where the filters return no results when you are in welsh. They work OK in english. Would you be able to have a look?

ChrisS1512 (changes requested):

src/main/resources/welshJurisdictionLookup.json
KianKwa: Removed welshJurisdictionLookup and added it to listLookup

src/main/views/subscription-configure-list.njk
KianKwa: `the issue should be fixed now i latest commit. Thanks form testing it :)

ChrisS1512 (approved):

PR #546, opened 2022-10-07
PUB-1552 Session idle auto sign out

KianKwa:
### JIRA link ###

Implement session idle auto sign out.

Screen when user is warned about auto signed out (20 mins before the expiry date):
image

Screen after user is auto signed out (after 1 hour of inactivity for media user and 4 hours for admin user):
image

Screen after user clicks the 'sign out' button
image

### Change description ###

https://tools.hmcts.net/jira/browse/PUB-1552

```
[ ] Yes
[x] No
```


src/main/service/sessionManagementService.ts
```diff
@@ -1,63 +1,49 @@
...
-      this.logOut(req, res, false);
-      return true;
-    }
-    return false;
-  }
+  private logOutRedirectUrl(isAdmin: boolean, adminWrongFlow: boolean, isSessionExpired: boolean, language: string): string {
+    const url = new URL(`${FRONTEND_URL}/${this.getRedirectionPath(adminWrongFlow, isSessionExpired)}`);
+    url.searchParams.append('lng', language);
 
-  private isSessionExpired(req): boolean {
```
ChrisS1512, line 35: I'm wondering if we still want this in

If someone doesn't have JS enabled then the frontend triggers won't hit so I think we still need something to catch it if the session expires

ChrisS1512 (changes requested):
Really cool piece of functionality :) Nice job

src/main/service/sessionManagementService.ts
KianKwa: You have a good point here Chris. If someone has JS enabled this bit of code wont hit before the user would have already been auto signed out. So this bit of code is only going to matter for those without JS so the got signed out too if they navigate somewhere else after the session expired. Have added it back in :)

ChrisS1512 (dismissed):

furnivall (dismissed):

ChrisS1512 (approved):

PR #548, opened 2022-10-11
PUB-1468 Care Standards List template

junaidiqbalmoj:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1468

### Change description ###

Template for Care Standards List.

Image 11-10-2022 at 09 17

[ ] Yes
[x] No


src/main/views/case-standards-list.njk
```diff
@@ -0,0 +1,104 @@
...
+          <th scope="col" class="govuk-table__header">{{ caseName }}</th>
+          <th scope="col" class="govuk-table__header">{{ duration }} </th>
+          <th scope="col" class="govuk-table__header">{{ hearingType }}</th>
+          <th scope="col" class="govuk-table__header">{{ venue }}</th>
+        </tr>
+        </thead>
+        <tbody class="govuk-table__body">
+        {% for courtCase in listData %}
+          {% set duration = '' %}
+          {% if courtCase['durationAsDays'] !== 0 %}
```
ChrisS1512, line 49: Wondering if this duration calculation should be moved into the JS (similar to 1373)

ChrisS1512 (changes requested):

src/main/views/case-standards-list.njk
```diff
@@ -0,0 +1,104 @@
...
+              {% elif durationAsHours === 0 %}
+              {% set duration = durationAsMinutes %}
+              {% elif durationAsMinutes === 0 %}
+              {% set duration = durationAsHours %}
+            {% endif %}
+          {% endif %}
+          <tr class="govuk-table__row">
+            <td class="govuk-table__cell">{{ courtCase['hearingDate'] }}</td>
+            <td class="govuk-table__cell">{{ courtCase['caseName'] }}</td>
+            <td class="govuk-table__cell">{{ duration }} {{ courtCase['caseSequenceIndicator'] }}</td>
```
KianKwa, line 85: I have seen in other lists, the caseSequenceIndicator hasn't got any square brackets in the data, and the square brackets are added in the nunjucks file. Are you sure it is different for the care standards list?

src/main/routes/routes.ts
```diff
@@ -67,6 +67,7 @@ export default function(app: Application): void {
   app.get('/alphabetical-search', app.locals.container.cradle.alphabeticalSearchController.get);
   app.post('/alphabetical-search', app.locals.container.cradle.alphabeticalSearchController.post);
   // app.get('/case-event-glossary', app.locals.container.cradle.caseEventGlossaryController.get);
+  app.get('/case-standards-list', app.locals.container.cradle.primaryHealthListController.get);
```
KianKwa, line 70: I think if you want to reuse the controller for primary health list, you should at least change the name of the controller so it is generic too

src/main/views/case-standards-list.njk
junaidiqbalmoj: Done. While doing this ticket, I though that we may be refactor this code as part of the refactoring ticket which you created other day, but I have done it for this list. Thanks.

src/main/views/case-standards-list.njk
junaidiqbalmoj: caseSequenceIndicator on frontend will be without brackets. Brackets must be part of Json. We should receive caseSequenceIndicator with brackets in json i.e. [1 of 2].

src/main/routes/routes.ts
junaidiqbalmoj: Agreed. Renamed Controller and Service.

ChrisS1512 (approved):
Looks good :) 

Build is failing but this looks to be a generic pipeline issue with chrome drivers so can fix seperately

KianKwa (approved):

PR #564, opened 2022-11-07
PUB-1385 Crime Magistrate Standard List

junaidiqbalmoj:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1385

### Change description ###

Template for Crime Magistrate Standard List which is Private list.

[ ] Yes
[x] No


src/main/service/listManipulation/magsStandardListService.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  private formatCaseTime(sitting: object, format: string): void {
+    if (sitting['sittingStart'] !== '') {
+      sitting['time'] = moment.utc(sitting['sittingStart']).tz(dataManipulationService.timeZone).format(format);
+    }
+  }
+
+  private manipulatePartyInformation(hearing: any, party: any): void {
+    let defendant = '';
+    switch (DataManipulationService.convertPartyRole(party.partyRole)) {
```
KianKwa, line 54: Swicth is normally used for 3 cases or more. Can we use if statement here instead?

src/main/service/listManipulation/magsStandardListService.ts
```diff
@@ -0,0 +1,92 @@
...
+    magsStandardListData['courtLists'].forEach(courtList => {
+      courtList['courtHouse']['courtRoom'].forEach(courtRoom => {
+        courtRoom['session'].forEach(session => {
+          session['sittings'].forEach(sitting => {
+            const allHearings = [];
+            session['formattedJudiciaries'] = dataManipulationService.findAndManipulateJudiciary(sitting);
+            this.formatCaseTime(sitting, 'h:mma');
+            hearingCount = hearingCount + sitting['hearing'].length;
+            sitting['hearing'].forEach(hearing => {
+              sitting['formattedDuration'] = dateTimeHelper.formatDuration(sitting['durationAsDays'] as number,
```
KianKwa, line 20: The method formatDuration() does not need anything from 'hearing'. May be it could be moved to the line before 'sitting['hearing].forEach' so it does not need to recalculated for each hearing.

src/main/service/listManipulation/magsStandardListService.ts
```diff
@@ -0,0 +1,92 @@
...
+                sitting['durationAsHours'] as number, sitting['durationAsMinutes'] as number, language, languageFile);
+              hearing.party.forEach(party => {
+                const hearingString = JSON.stringify(hearing);
+                const hearingObject = JSON.parse(hearingString);
+                if (hearingObject?.party) {
```
KianKwa, line 25: Hhmm why do we need the if statement here? It is within the forEach for party so surely it has party?

src/main/service/listManipulation/magsStandardListService.ts
junaidiqbalmoj: fixed.

src/main/service/listManipulation/magsStandardListService.ts
junaidiqbalmoj: fixed.

src/main/service/listManipulation/magsStandardListService.ts
junaidiqbalmoj: fixed.

KianKwa (dismissed):

src/main/service/listManipulation/magsStandardListService.ts
```diff
@@ -0,0 +1,90 @@
...
+            const allHearings = [];
+            session['formattedJudiciaries'] = dataManipulationService.findAndManipulateJudiciary(sitting);
+            this.formatCaseTime(sitting, 'h:mma');
+            sitting['formattedDuration'] = dateTimeHelper.formatDuration(sitting['durationAsDays'] as number,
+              sitting['durationAsHours'] as number, sitting['durationAsMinutes'] as number, language, languageFile);
+            hearingCount = hearingCount + sitting['hearing'].length;
+            sitting['hearing'].forEach(hearing => {
+              if (hearing?.party) {
+                hearing.party.forEach(party => {
+                  const hearingString = JSON.stringify(hearing);
```
ChrisS1512, line 24: I'm wondering if we can remove lines 24/25 (and instead directly just use hearing)

src/main/service/listManipulation/magsStandardListService.ts
```diff
@@ -0,0 +1,90 @@
...
+                  });
+                  allHearings.push(hearingObject);
+                });
+              }
+            });
+            sitting['hearing'] = allHearings;
+          });
+        });
+        courtRoom['totalHearing'] = hearingCount;
+        hearingCount = 0;
```
ChrisS1512, line 39: Rather than setting hearingCount back to 0 could the let on line 10 be moved into one of the loops instead?

ChrisS1512 (changes requested):

src/main/service/listManipulation/magsStandardListService.ts
junaidiqbalmoj: We are not using hearingCount for this list. redundant code :) removed.

joshblackmoor (approved):

src/main/views/magistrates-standard-list.njk
```diff
@@ -0,0 +1,216 @@
...
+        {% for sitting in session['sittings'] %}
+          {% for hearing in sitting['hearing'] %}
+            {% if hearing['defendantHeading'] | length %}
+              {% for case in hearing['case'] %}
+                  <div class="govuk-accordion__section govuk-accordion__section--expanded">
+                    <div class="govuk-accordion__section-header">
+                      <h2 class="govuk-accordion__section-heading">
+                      <span class="govuk-accordion__section-button" id="accordion-default-heading-{{ sessionCount }}">
+                        {% if hearing['gender'] | length %}
+                          {{ defendantNameHeading }} {{ hearing['defendantHeading'] }} ({{ hearing['gender'] }})
```
ChrisS1512, line 102: I think we are missing an 'In Custody' flag here after the gender

ChrisS1512 (changes requested):

src/main/views/magistrates-standard-list.njk
junaidiqbalmoj: ok, so that's what asterisk means :) thanks. Added

ChrisS1512 (dismissed):

KianKwa (approved):

ChrisS1512 (approved):

KianKwa (approved):

PR #567, opened 2022-11-09
PUB 1382 firm list template

furnivall:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-1382

### Change description ###
Added Crown Firm List. This is the first time we've used the proper nunjucks method for the accordions so am also adding a ticket to refactor other lists to this in future.

image

```
[ ] Yes
[x] No
```


src/main/service/listManipulation/crownFirmListService.ts
```diff
@@ -0,0 +1,114 @@
...
+        const uniqueCourtRooms = dataManipulationService.uniquesInArrayByAttrib(record, 'courtRoom');
+        Array.from(uniqueCourtRooms).forEach(courtRoom => {
+          const room = record.filter(row => row.courtRoom === courtRoom);
+          thisDayCourts.push({'courtRoom': courtRoom, data: room});
+        });
+        // custom sort def - basically if it's got the string, move to end. Needed to be suppressed because eslint
+        // does not understand that a compare function needs two vars.
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        function compare (a, b) {
+          if (a.courtRoom.toLowerCase().includes('to be allocated')){return 1;} return -1;}
```
KianKwa, line 104: The text "to be allocated" can come in both direction. You are comparing just one direction here. So any text which is not "to be allocated" will get bumped up as you return '-1'. So you are likely to have the order of the array reversed for all other texts. For example if you have an array ['a', 'b', 'to be allocated', 'c', 'd'], the result will be ['d', 'c', 'b', 'a', 'to be allocated']. 

It is fine if that is what you intend to do. If you want to enforce the order ['a', 'b', 'c', 'd', 'to be allocated'] might be better to consider both directions with a compare method like below:

function compare (a, b) {
      if (a.toLowerCase().includes('to be allocated')) {
        return 1;
      }
      if (b.toLowerCase().includes('to be allocated')) {
        return -1;
      }
      return 0;
    }

src/main/service/listManipulation/crownFirmListService.ts
```diff
@@ -0,0 +1,114 @@
...
+                  courtRoom: courtRoom['courtRoomName'],
+                  joh: session['formattedJudiciaries'],
+                  durationAsHours: sitting['durationAsHours'],
+                  durationAsMinutes: sitting['durationAsMinutes'],
+                  formattedDuration: sitting['formattedDuration'],
+                  caseNumber: thisCase['caseNumber'],
+                  caseSeparator: thisCase['caseSequenceIndicator'],
+                  linkedCases: thisCase['linkedCases'],
+                  hearingNotes: hearing['listingNotes'],
+                  defendant: formattedName[1].toUpperCase() + ', ' + formattedName[0],
```
KianKwa, line 48: I am assuming formattedName[1] is always surname and  formattedName[0] is always forename? If these fields can be empty, we are likely to see a redundant comma and the output could be for example:

Furnivall,
, Daniel

...or just a single comma. Might be better to check both surname and forename are present before adding the comma?

src/main/controllers/CrownFirmListController.ts
```diff
@@ -0,0 +1,46 @@
...
+import { LocationService } from '../service/locationService';
+import { DataManipulationService } from '../service/dataManipulationService';
+import {CrownFirmListService} from '../service/listManipulation/crownFirmListService';
+
+const publicationService = new PublicationService();
+const locationService = new LocationService();
+const dataManipulationService = new DataManipulationService();
+const firmListService = new CrownFirmListService();
+
+export default class CrownFirmListController {
```
KianKwa, line 15: THink we are missing the unit test for the controller. Could you add that in please? :)

src/main/service/listManipulation/crownFirmListService.ts
furnivall: Yeah, I was gonna do this at initial implementation but i just thought it wasn't really a requirement. Can add now.

src/main/views/crown-firm-list.njk
```diff
@@ -0,0 +1,171 @@
...
+            <li>{{ restrictionBulletPoint2Bill }}</li>
+          </ul>
+          <p class="govuk-body">{{ restrictionInformationP5Bill }} <a href="#"
+                                                                      class="govuk-link">{{ restrictionLinkBill }}</a>
+          </p>
+        </div>
+      {% endif %}
+    </div>
+  </div>
+  {% for court in allocated %}
```
ChrisS1512, line 83: Only super minor, but there's a few formatting differences below (some blocks have spaces before, some don't etc)

Could this be updated to make it consistent?

ChrisS1512 (changes requested):

src/main/service/listManipulation/crownFirmListService.ts
KianKwa: ok if we don't care about the ordering it probably doesn't matter but having the second comparison part will at least maintain the original order for the other elements.

src/main/service/listManipulation/crownFirmListService.ts
furnivall: Fair point! Fixed.

src/main/views/crown-firm-list.njk
furnivall: sorted I think?

src/main/controllers/CrownFirmListController.ts
furnivall: now done :)

src/main/service/listManipulation/crownFirmListService.ts
furnivall: added btw

ChrisS1512 (dismissed):

src/main/views/crown-firm-list.njk
```diff
@@ -0,0 +1,162 @@
...
+{% endblock %}
+{% block content %}
+  <div class="parent-box">
+    <h2 id='page-heading' class="govuk-heading-l">
+      {{ courtName }}: {{ headerPart1 }}
+      {% if bill %}
+        <br/>/{{ courtName }}: {{ headerPart2 }}
+      {% endif %}
+    </h2>
+    <p class="govuk-body govuk-!-font-weight-bold">{{ listDate }} {{ contentDate }}</p>
```
junaidiqbalmoj, line 22: As per mockup, List for has two dates like 25 May 2022 to 30 May 2022

src/main/resources/locales/en/crown-firm-list.json
```diff
@@ -0,0 +1,36 @@
+{
+  "title": "Crown Firm List",
+  "headerPart1": "Firm List",
+  "listDate": "List for",
+  "listUpdated": "Last Updated DATE at ",
```
junaidiqbalmoj, line 5: Missing colon after last updated.

src/main/resources/locales/cy/crown-firm-list.json
```diff
@@ -0,0 +1,36 @@
+{
+  "title": "",
+  "headerPart1": "",
+  "listDate": "Rhestr ar gyfer",
+  "listUpdated": "Diweddarwyd diwethaf DATE yn ",
```
junaidiqbalmoj, line 5: Missing colon after Diweddarwyd diwethaf.

src/main/resources/locales/bill/crown-firm-list.json
```diff
@@ -0,0 +1,47 @@
+{
+  "title": "",
+  "headerPart1": "",
+  "headerPart2": "Rhestr Ddyddiol",
+  "listDate": "List for/Rhestr ar gyfer",
+  "listUpdated": "Last Updated/Diweddarwyd diwethaf DATE at/yn ",
```
junaidiqbalmoj, line 6: Missing colon after Last Updated/Diweddarwyd

src/main/service/listManipulation/crownFirmListService.ts
```diff
@@ -0,0 +1,116 @@
...
+            sitting['hearing'].forEach(hearing => {
+              dailyListService.findLinkedCasesInformation(hearing);
+              hearing['case'].forEach(thisCase => {
+                const formattedName = hearing['defendant'].split(/\s+/);
+                const row = {
+                  courtName: courtName,
+                  sittingDate: sittingDate,
+                  sittingTime: dataManipulationService.publicationTimeInBst(sitting['sittingStart']),
+                  courtRoom: courtRoom['courtRoomName'],
+                  joh: session['formattedJudiciaries'],
```
junaidiqbalmoj, line 40: formattedJudiciaries is not working as expected. If there is Judiciaries section in SITTING but not in SESSION, it will ignore the SITTING Judiciaries data. This is bug which I fixed in Mags Standard list. 
You can fix it:
In SittingLoop, do something like
`session['sittings'].forEach(sitting => {
            const judiciary = dataManipulationService.findAndManipulateJudiciary(sitting);
            if (judiciary !== '') {
              session['formattedJudiciaries'] = judiciary;
            }`

junaidiqbalmoj (changes requested):
Formatting of table is not align with mockup. We use to have horizontal scroll bar for the table which is missing in this list. 

error

src/main/service/listManipulation/crownFirmListService.ts
furnivall: > Formatting of table is not align with mockup. We use to have horizontal scroll bar for the table which is missing in this list.
> 
> error

Sorted :)

src/main/resources/locales/bill/crown-firm-list.json
furnivall: This has now been fixed.

src/main/resources/locales/en/crown-firm-list.json
furnivall: This has now been fixed.

KianKwa (approved):

junaidiqbalmoj (dismissed):

ChrisS1512 (approved):

junaidiqbalmoj (approved):

PR #574, opened 2022-11-16
PUB-1613 System Admin User management dashboard

joshblackmoor:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1613

### Change description ###
Added new system admin user management pages to: Find a user, update their role and delete them. 

Note: E2Es will fail until Account management PR is merged. 

```
[ ] Yes
[x] No
```


src/main/resources/requests/accountManagementRequests.ts
```diff
@@ -187,4 +187,74 @@ export class AccountManagementRequests {
...
+  public async deleteUser(userId: string, adminUserId: string): Promise<object> {
+    try {
+      logger.info('User with ID: ' + userId + ' deleted by Admin with ID: ' + adminUserId);
+      const response = await accountManagementApi.delete(`/account/delete/${userId}`);
+      return response.data;
+    } catch (error) {
+      if (error.response) {
+        console.log(error.response.data);
+      } else if (error.request) {
+        console.log(JSON.stringify(error.request));
```
KianKwa, line 234: I am guessing this is was used for local testing only? If so we don't need it anymore.

src/main/resources/requests/accountManagementRequests.ts
```diff
@@ -187,4 +187,74 @@ export class AccountManagementRequests {
...
+  public async updateUser(userId: string, role: string, adminUserId: string): Promise<object> {
+    try {
+      logger.info('User with ID: ' + userId + ' role updated to ' + role + ' by Admin with ID: ' + adminUserId);
+      const response = await accountManagementApi.put(`/account/update/${userId}/${role}`);
+      return response.data;
+    } catch (error) {
+      if (error.response) {
+        console.log(error.response.data);
+      } else if (error.request) {
+        console.log(JSON.stringify(error.request));
```
KianKwa, line 252: Same as the comment before :)

src/main/resources/requests/accountManagementRequests.ts
joshblackmoor: Lol yeah, removed it :)

src/main/service/userManagementService.ts
```diff
@@ -0,0 +1,330 @@
...
+      rowObject['actions'] = {items};
+    }
+    return rowObject;
+  }
+
+  /**
+   * Format and return the date with the correct format.
+   */
+  private formatDate(rawDate: any) {
+    return moment(rawDate).format('DD/MM/YYYY HH:mm:ss');
```
KianKwa, line 309: I am not sure the time is going to be formatted correctly during daylight saving. It will be formatted correctly after 30 October this year but if the datetime is during daylight saving, it will be one hour off. The display time should be in the local UK time regardless of whether it is GMT or BST.

In account management, we have the following line which indicates the datetime is stored in the UTC format:

LocalDateTime localDateTime = LocalDateTime.now(ZoneId.of("UTC"));

When the date time is received from Account management in front-end it lost the 'Z' suffix the UTC indicator because converting a LocalDateTime to a string does not take into account the time format during serialisation. I have this PR here fixing that for data management. Think we will need to do the same for account management timestamp.

https://github.com/hmcts/pip-data-management/pull/208/files#diff-3830eb72074ec4e3ff516b380632461e8a87e5272503a024ebcc4591d5b0cb6a

After that we need to format the time in front end using the following line and that should format the time correctly.

moment.utc(rawDate).tz('europe/london').format('DD/MM/YYYY HH:mm:ss');

joshblackmoor: Good spot. Have updated this now and also in the account management PR :)

KianKwa (dismissed):

ChrisS1512 (dismissed):

ChrisS1512 (approved):

KianKwa (approved):

PR #584, opened 2022-11-22
1630 - List Template Refactoring

furnivall:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PUB-1630

### Change description ###
Moving service layer for the following list types out of the dataManipulationService class:
- SJP, SSCS, ET, COP and IAC

```
[ ] Yes
[ x] No
```


src/main/modules/nunjucks/njkFilters.ts
```diff
@@ -0,0 +1,47 @@
+import path from 'path';
+import moment from 'moment';
+import { PublicationService } from '../../service/publicationService';
+import { printableDuration } from './printableDuration';
+const publicationService = new PublicationService();
+
+function createFilters (env) {
```
ChrisS1512, line 7: Really nice :)

ChrisS1512 (changes requested):

src/main/service/listParseHelperService.ts
```diff
@@ -0,0 +1,359 @@
...
+
+export class ListParseHelperService {
+
+  public timeZone = 'Europe/London';
+
+  /**
+   * Manipulate the daily cause list json data for writing out on screen.
+   * @param dailyCauseList
+   */
+  public manipulatedDailyListData(dailyCauseList: string): object {
```
KianKwa, line 12: This needs to be renamed to somethibg more generic.....I see it is used by public list and standard list too, so if possible don't use the word 'DailyLIst' in the method name

src/main/service/listParseHelperService.ts
```diff
@@ -0,0 +1,359 @@
...
+      let durationAsDays = 0;
+      if(durationAsHours >= 24) {
+        durationAsDays = Math.floor(durationAsHours / 24);
+      }
+
+      sitting['durationAsHours'] = durationAsHours;
+      sitting['durationAsMinutes'] = durationAsMinutes;
+      sitting['durationAsDays'] = durationAsDays;
+
+      sitting['time'] = moment.utc(sitting['sittingStart']).tz(this.timeZone).format('HH:mm');
```
KianKwa, line 276: I wonder line 276 and 282 should be moved out to a different method. It has got nothing to do with duration and shoehorned into this method. Also some of the lists using 'time', and some lists using 'startTime' but all for the same thing only different format. If you do that will need to find out what lists using these 2 fields and they need to call this method individually. So this could be a lot more work for u. We can log it and fix this issue separately so it is up to u.

src/main/service/listManipulation/CrimeListsService.ts
```diff
@@ -1,8 +1,8 @@
 import moment from 'moment-timezone';
```
KianKwa, line 1: Somewhere in this case, there is a method called "findUnallocatedCasesInCrownDailyListData" which is used by other lists too. This doesnt sound very generic so consider renaming it

KianKwa: This seems to be used by CrownDailyListController only. If so could just move it to the daily list controller class. Strange that I can see some tests for other lists using that too. Need to find out more

KianKwa, line 1: A few of the methods in the CrimeListService are used by magistrate list too. I wonder should rename the class to CrownListService and not CrimeListService so the name of the class reflects its usage. What do u think?

src/main/service/listParseHelperService.ts
```diff
@@ -0,0 +1,359 @@
...
+        sitting['startTime'] = moment.utc(sitting['sittingStart']).tz(this.timeZone).format('h.mma');
+      }
+    }
+  }
+
+  /**
+   * Function which extracts the time from a UTC Date Time in BST format.
+   * @param publicationDatetime The publication date time to convert in UTC.
+   */
+  public publicationTimeInBst(publicationDatetime: string): string {
```
KianKwa, line 290: The terminology used for this method and publicationDateInBst is not correctly. Wonder we should update them? What these methods doing are update the time to UK Local time and not just BST. Currently we are not in daylight saving but in GMT time and this method is returning datetime in GMT now. In the summer it is right that we are retuning BST time. May be can rename it to publicationTimeInUKLocalTime or something else more appropriate?

src/main/service/listManipulation/CrimeListsService.ts
furnivall: Gonna leave this as is for this round of refactoring - I don't feel all that comfortable with funcs in controller that aren't request types.

KianKwa (dismissed):

KianKwa (approved):

ChrisS1512 (dismissed):

junaidiqbalmoj (approved):

KianKwa (approved):

PR #606, opened 2022-11-22
PUB-1677 Media ID bug

joshblackmoor:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1677

### Change description ###

Converted the extension to lowercase when looking up the mapping as before JPG and PNG failed.

```
[ ] Yes
[x] No
```


KianKwa (dismissed):

src/main/controllers/MediaAccountReviewController.ts
```diff
@@ -33,7 +33,7 @@ export default class MediaAccountReviewController {
       const imageName = applicant.imageName;
       const extension = imageName.substring(imageName.lastIndexOf('.') + 1, imageName.length);
 
-      const contentType = allowedImageTypeMappings[extension];
+      const contentType = allowedImageTypeMappings[extension.toLowerCase()];
```
ChrisS1512, line 36: Could we add a small unit test just checking this? Will act as a regression test for the future

ChrisS1512 (changes requested):

src/main/controllers/MediaAccountReviewController.ts
joshblackmoor: Added tests :)

KianKwa (approved):

ChrisS1512 (approved):

PR #609, opened 2022-11-23
PUB-1533 - Cancelled password reset

ChrisS1512:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1533

### Change description ###

Added cancelled password reset page

image

```
[ ] Yes
[x] No
```


joshblackmoor (changes requested):
Code looks good. I think we need to update the password reset URLs in flux config that we send with notify to reflect this new change.

ChrisS1512:
> Code looks good. I think we need to update the password reset URLs in flux config that we send with notify to reflect this new change.

Good spot! I've added PR's for sds flux + pub services

KianKwa (approved):

joshblackmoor (approved):

joshblackmoor (approved):

KianKwa (approved):

PR #614, opened 2022-11-24
PUB-1648 Daily cause list clean up

joshblackmoor:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1648

### Change description ###

Cleaned up daily cause list as per ticket.

```
[ ] Yes
[x] No
```


src/main/service/dataManipulationService.ts
```diff
@@ -542,12 +542,12 @@ export class DataManipulationService {
       sitting['durationAsHours'] = durationAsHours;
       sitting['durationAsMinutes'] = durationAsMinutes;
       sitting['durationAsDays'] = durationAsDays;
-
-      sitting['time'] = moment.utc(sitting['sittingStart']).tz(this.timeZone).format('HH:mm');
       const min = moment(sitting['sittingStart'], 'HH:mm').minutes();
       if (min === 0) {
```
ChrisS1512, line 546: It looks like startTime is identical to time, could this be used instead?

ChrisS1512 (changes requested):

src/main/service/dataManipulationService.ts
joshblackmoor: Fixed as discussed on slack

ChrisS1512 (approved):

furnivall (approved):

furnivall (approved):

PR #615, opened 2022-11-28
PUB-1614 System admin bulk create media accounts

KianKwa:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1614

### Change description ###

System admin bulk create media accounts

```
[ ] Yes
[x] No
```


src/main/controllers/BulkCreateMediaAccountsConfirmationController.ts
```diff
@@ -0,0 +1,79 @@
...
+const bulkCreateAccountsConfirmationUrl = 'bulk-create-media-accounts-confirmation';
+const bulkCreateAccountsConfirmedUrl = 'bulk-create-media-accounts-confirmed';
+
+export default class BulkCreateMediaAccountsConfirmationController {
+  public async get(req: PipRequest, res: Response): Promise<void> {
+    const formData = (req.cookies?.formCookie) ? JSON.parse(req.cookies['formCookie']) : {};
+    const fileName = formData['uploadFileName'];
+
+    const accountsToCreate =
+      fileName === undefined ? []
```
ChrisS1512, line 20: This scenario should never happen in normal operation, so I wonder if this should go to an error page if it's undefined?

src/main/controllers/BulkCreateMediaAccountsConfirmationController.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }
+
+  private static async getAccountsToCreate(userId, fileName): Promise<string[][]> {
+    const file = await fileHandlingService.readFileFromRedis(userId, fileName);
+    const fileData = fileHandlingService.readCsvToArray(file);
+    return BulkCreateMediaAccountsConfirmationController.reorderFieldsInRows(fileData);
+  }
+
+  private static reorderFieldsInRows(rows): string[][] {
```
ChrisS1512, line 66: The backend doesn't need a specific order, so wondering if we can remove this below?

ChrisS1512 (changes requested):

src/main/controllers/BulkCreateMediaAccountsConfirmationController.ts
KianKwa: There is a table on that confirmation screen showing the list of media accounts to be created. We need the ordering here so the table header matches the table contents :)

furnivall (approved):

furnivall (dismissed):

ChrisS1512 (approved):

joshblackmoor (approved):

PR #657, opened 2022-12-15
PUB-1628 Luxon DateTime library

junaidiqbalmoj:
### JIRA link ###

https://tools.hmcts.net/jira/browse/PUB-1628

### Change description ###

Stop using moment.js library and start using using Luxon.js.

[ ] Yes
[x] No


src/main/controllers/SjpPressListController.ts
```diff
@@ -32,7 +32,7 @@ export default class SjpPressListController {
         sjpData: manipulatedData,
         publishedDateTime: publishedDate,
         publishedTime: publishedTime,
-        contactDate: moment.utc(Date.parse(metaData['contentDate'])).format('D MMMM YYYY'),
+        contactDate: DateTime.fromISO(metaData['contentDate'], {zone: 'utc'}).toFormat('d MMMM yyyy'),
```
ChrisS1512, line 35: I've noticed we're using zone: 'utc' here and zone: 'europe/london' in some of the other lists. We should use the same accross all lists that converts the UTC to BST

ChrisS1512: Just looking further, would the helperService.contentDateInUtcTime work for all lists? Potentially could switch to that

src/main/service/listParseHelperService.ts
```diff
@@ -250,26 +251,35 @@ export class ListParseHelperService {
...
-        sitting['time'] = moment.utc(sitting['sittingStart']).tz(this.timeZone).format('h:mma');
+        this.formatCaseTime(sitting, 'h:mma');
       }
     }
   }
 
+  public formatCaseTime(sitting: object, format: string): void {
+    if (sitting['sittingStart'] !== '') {
+      let zonedDateTime = DateTime.fromISO(sitting['sittingStart'], {zone: this.timeZone});
+      //Unable to find a way to convert time to BST. Luxon always return offset (+01:00) with the time,
```
ChrisS1512, line 265: I've tried out a few lists for the Content Date, and it seems to work OK (In summer, adds one hour) with the normal DateTime.fromISO.

Wondering if the same could be done in this method rather than having to manually apply the offset

src/main/controllers/SjpPressListController.ts
junaidiqbalmoj: As discussed, I added Welsh to all the lists.

src/main/service/listParseHelperService.ts
junaidiqbalmoj: As discussed, I am checking the Z at the end of time, if it is there, don't add offset to the time.

ChrisS1512 (approved):

junaidiqbalmoj (commented):
Thanks @joshblackmoor for pointing out the bug. It's fixed.
image

joshblackmoor (dismissed):

package.json
```diff
@@ -71,6 +71,7 @@
     "jquery": "^3.6.1",
     "jwt-decode": "^3.1.2",
     "lodash": "^4.17.21",
+    "luxon": "^3.1.1",
     "mark.js": "^8.11.1",
     "minimist": "^1.2.7",
     "moment": "2.29.4",
```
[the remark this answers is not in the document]
junaidiqbalmoj: it's not needed. Removed.

joshblackmoor (approved):

ChrisS1512 (approved):

PR #777, opened 2023-02-07
PUB-1852 - Selecting sensitivity

ChrisS1512:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-1852

### Change description

Adds a number of changes to the manual upload process, including:

- List types can now have a default sensitivity
- If a list type has no default sensitivity, the sensitivity box will be blank and therefore must be selected
- If the user doesn't select a sensitivity, an error will occur
- If the selected sensitivity is different from the default sensitivity, a warning will display on the confirm page
- Available to renamed to sensitivity

image

```
[ ] Yes
[x] No
```


src/main/service/manualUploadService.ts
```diff
@@ -72,8 +102,9 @@ export class ManualUploadService {
                 language,
                 languageFile
             ),
+            classificationError: formValues['classification'] ? null : 'true',
         };
-        if (!fields.courtError && !fields.contentDateError && !fields.displayDateError) {
+        if (!fields.courtError && !fields.contentDateError && !fields.displayDateError && !fields.classificationError) {
```
KianKwa, line 107: Lots of negative conditions here. Is that better to do the following so it will stop evaluating as soon as if find an error?
if (fields.courtError || fields.contentDateError || fields.displayDateError || fields.classificationError) {
    return fields;
}
return null;

ChrisS1512: Good spot. I've updated this now.

KianKwa (dismissed):

joshblackmoor (dismissed):

junaidiqbalmoj (dismissed):

KianKwa (dismissed):

joshblackmoor (approved):

KianKwa (approved):

PR #823, opened 2023-03-02
PUB-1807 - Changing of own account

ChrisS1512:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-1807

### Change description

Updated so that when a user attempts to update their own account, an error message is displayed.

image

```
[ ] Yes
[x] No
```


KianKwa (approved):

src/test/end-to-end/tests/change-own-user-test.ts
```diff
@@ -0,0 +1,13 @@
+import { config as testConfig } from '../../config';
+
+Feature('Change own user role');
+
+Scenario('I as an admin should be able to see an error message if I try to change my own user', async ({ I }) => {
+    I.loginAsAdmin();
+    I.click('#card-admin-management');
+    I.fillField('#search-input', testConfig.ADMIN_USERNAME as string);
```
furnivall, line 8: should be a secret wrapped thing right?

furnivall (commented):
Really nice - I like the logging of the attempts!
Will approve once change is made :)

src/test/end-to-end/tests/change-own-user-test.ts
ChrisS1512: Good spot! I've updated this

KianKwa (approved):

PR #833, opened 2023-03-10
PUB-1742 - Fix code smells

ChrisS1512:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-1742

### Change description

Changes to update:

- Code Smells
- Improve test coverage
- Remove uneeded conditionals

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/test/unit/controllers/UserManagementController.test.ts
+++ b/src/test/unit/controllers/UserManagementController.test.ts
@@ -25,10 +25,9 @@ const formattedDataStub = sinon.stub(UserManagementService.prototype, 'getFormat
 
 formattedDataStub
     .withArgs(
-        new UserSearchCriteria(parseInt(testPage), testEmail, testUserId, testProvenanceId, testRoles, testProvenances),
+        new UserSearchCriteria(testPage, testEmail, testUserId, testProvenanceId, testRoles, testProvenances),
         testPath,
-        userId,
-        email
+        { userId: userId, email: email }
     )
     .returns({
         userData: 'otherData',


src/main/controllers/UserManagementController.ts
```diff
@@ -14,12 +15,14 @@ export default class UserManagementController {
...
         } else {
             const pageData = await userManagementService.getFormattedData(
-                parseInt(<string>req.query.page) || 1,
-                (req.query?.email as string) || '',
-                (req.query?.userId as string) || '',
-                (req.query?.userProvenanceId as string) || '',
-                (req.query?.roles as string) || '',
-                (req.query?.provenances as string) || '',
+                new UserSearchCriteria(
+                    req.query.page,
```
KianKwa, line 19: "req.query" is not mark as optional here but they are flagged as optional in the subsequent parameters. Can make them consistent

src/main/service/listManipulation/MagistratesStandardListService.ts
```diff
@@ -83,6 +39,56 @@ export class MagistratesStandardListService {
         return magsStandardListData;
     }
 
+    private processHearing(sitting, language) {
+        const allDefendants = [];
+        sitting['hearing'].forEach(hearing => {
+            hearing?.party?.forEach(party => {
```
KianKwa, line 45: We are now inside the forEach loop for each hearing element, so the hearing is not optional anymore. Can remove the question mark after hearing.

src/main/service/listManipulation/CrimeListsService.ts
```diff
@@ -91,18 +92,16 @@ export class CrimeListsService {
...
-                        linkedCases += caseLinked.caseId.trim();
-                        linkedCases += helperService.stringDelimiter(linkedCases?.length, ',');
-                    });
-                }
-                cases['linkedCases'] = linkedCases?.replace(/,\s*$/, '').trim();
-            });
-        }
+        hearing?.case.forEach(cases => {
+            linkedCases = '';
+            if (cases?.caseLinked) {
```
KianKwa, line 97: The question mark can be removed to because we are inside the forEach loop of the case element

src/main/controllers/UserManagementController.ts
ChrisS1512: Good spot, now marked as optional :)

src/main/service/listManipulation/MagistratesStandardListService.ts
ChrisS1512: Question mark removed

src/main/service/listManipulation/CrimeListsService.ts
ChrisS1512: Question mark removed

src/main/service/listManipulation/CrimeListsService.ts
```diff
@@ -91,18 +92,16 @@ export class CrimeListsService {
...
-                if (cases?.caseLinked) {
-                    cases.caseLinked.forEach(caseLinked => {
-                        linkedCases += caseLinked.caseId.trim();
-                        linkedCases += helperService.stringDelimiter(linkedCases?.length, ',');
-                    });
-                }
-                cases['linkedCases'] = linkedCases?.replace(/,\s*$/, '').trim();
-            });
-        }
+        hearing?.case.forEach(cases => {
```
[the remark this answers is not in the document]
ChrisS1512: I've checked all the lists that call this, and case is mandatory so I've kept that as it is.

I have removed the 'hearing' optional though, as it will always be present when calling this

KianKwa (dismissed):

joshblackmoor (dismissed):

KianKwa (approved):

joshblackmoor (approved):

PR #842, opened 2023-03-28
PUB-1957 Update error page

KianKwa:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-1957

### Change description

Update error page

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/locales/cy/error.json
+++ b/src/main/resources/locales/cy/error.json
@@ -1,4 +1,6 @@
 {
-    "header": "Mae rhywbeth wedi mynd o'i le",
+    "header": "Sorry, there is a problem with the service",
+    "retryMessage": "Please try again later.",
+    "contactUsMessage": "If the problem persists please contact our courts and tribunals service centre on 0300 303 0656.",
     "backButton": "Yn ôl"
 }
--- a/src/test/unit/views/summary-of-publications.test.ts
+++ b/src/test/unit/views/summary-of-publications.test.ts
@@ -11,6 +11,7 @@ courtStub.withArgs(10).resolves(JSON.parse('{"name":"New Court"}'));
 describe('Summary of publications page', () => {
     let htmlRes: Document;
 
+    const headingClass = 'govuk-heading-xl';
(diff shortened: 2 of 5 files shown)


src/main/resources/locales/cy/error.json
```diff
@@ -1,4 +1,6 @@
 {
-    "header": "Mae rhywbeth wedi mynd o'i le",
+    "header": "Sorry, there is a problem with the service",
+    "retryMessage": "Please try again later.",
```
ChrisS1512, line 3: These two lines end with full stops, however the ticket screenshot does not. Could you double check which one is needed?

src/test/unit/views/summary-of-publications.test.ts
```diff
@@ -67,4 +68,28 @@ describe('Summary of publications page', () => {
             expect(body[4].innerHTML).contains('Sorry, no lists found for this court', 'Non SJP is not displayed');
         });
     });
+
+    describe('Non SJP Summary of Pubs without locationId query parameter', () => {
```
ChrisS1512, line 72: I'm thinking we should create a separate error view test for this, as although using the summary of pubs to trigger it, it's specifically the error view

ChrisS1512 (changes requested):

src/main/resources/locales/cy/error.json
KianKwa: Chris C has confirmed we need the full stops. They are sentences within the body not heading.

src/test/unit/views/summary-of-publications.test.ts
KianKwa: I have moved these 2 tests out to error.test.ts. I am still using summary of publications page on the error test as this error page hasn't got its owned url.

ChrisS1512 (dismissed):

joshblackmoor (approved):

ChrisS1512 (approved):

joshblackmoor (approved):

PR #872, opened 2023-04-06
PUB-1971 Randomise test court ID and name fir E2E tests

KianKwa:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-1971

### Change description

Randomise test court ID and name fir E2E tests

```
[ ] Yes
[x] No
```


joshblackmoor (approved):

src/test/end-to-end/shared/shared-functions.ts
```diff
@@ -1,3 +1,63 @@
...
+            Provenance: 'ListAssist',
+            'Provenance Location ID': '3482',
+            'Provenance Location Type': 'Venue',
+        },
+    ];
+};
+
+const generateTestLocationFields = (): [number, string] => {
+    // Generate a random test court ID between 200 and 1000000
+    const id = randomData.getRandomNumber(200, 1000000);
```
ChrisS1512, line 27: I would up the starting number a bit just so it doesn't overlap when we get more courts.

ChrisS1512 (changes requested):

src/test/end-to-end/shared/shared-functions.ts
KianKwa: Now using a bigger starting number and a larger range for the court ID

ChrisS1512 (approved):

src/test/end-to-end/shared/shared-functions.ts
```diff
@@ -1,3 +1,63 @@
...
+            Provenance: 'ListAssist',
+            'Provenance Location ID': '3482',
+            'Provenance Location Type': 'Venue',
+        },
+    ];
+};
+
+const generateTestLocationFields = (): [number, string] => {
+    // Generate a random test court ID between 1000 and 9999999999
+    const id = randomData.getRandomNumber(1000, 9999999999);
```
ChrisS1512, line 27: The E2E test seems to be failing (ran a few times with same error). 

I think (but not checked the logs) it could be to do with the max int here. It's above the max java int (2147483647) so I think it's worth changing it to that and trying again

ChrisS1512 (changes requested):
The delete court test seems to be failing (ran a few times but same error).

src/test/end-to-end/shared/shared-functions.ts
KianKwa: Have removed one digit from the max number. Doesnt need to be that huge. It is passing now

ChrisS1512 (approved):

ashwini-mv (approved):

PR #926, opened 2023-05-18
Pub 1999

NatashaAlker:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-1999

### Change description

Display "ListAssist" rather than "SNL" as the Data Source when showing lists

```
[ ] Yes
[X] No
```


src/main/modules/nunjucks/njkFilters.ts
```diff
@@ -148,6 +148,10 @@ function createFilters(env) {
         }
         return result;
     });
+
+    env.addFilter('maskLegacyDataSource', function (provenance) {
+        return provenance == 'SNL' ? 'LIST_ASSIST' : provenance;
```
joshblackmoor, line 153: Same comment as on other PR if it should be ListAssist or LIST_ASSIST :)

joshblackmoor (commented):
Just wondering if we could add a test or two on one of the views to check that when its SNL it is successfully updated to the new name? Happy either way

src/main/controllers/CrownWarnedListController.ts
```diff
@@ -50,7 +49,7 @@ export default class CrownWarnedListController {
                 publishedDate: publishedDate,
                 publishedTime: publishedTime,
                 version: searchResults['document']['version'],
-                provenance: metaData.provenance,
+                provenance: metaData['provenance'],
```
ChrisS1512, line 52: I think this can be changed back to .provenance (It provides some type safety over the ['provenance'] version)

ChrisS1512 (changes requested):

ChrisS1512 (approved):

joshblackmoor (approved):
Nice :) Happy first ticket on pip

KianKwa (approved):

PR #958, opened 2023-06-02
Pub 1841

NatashaAlker:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-1841

### Change description

Functional tests for Admin Dashboard User Management

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/test/end-to-end/tests/admin/user-management-test.ts
+++ b/src/test/end-to-end/tests/admin/user-management-test.ts
@@ -1,36 +1,84 @@
-import {config as testConfig} from "../../../config";
+import { config as testConfig } from '../../../config';
+import {randomData} from "../../shared/random-data";
 
 Feature('Manage admin user accounts');
 
+const TEST_WORKER_NUMBER = randomData.getRandomNumber(10000000, 99999999);
 const TEST_FIRST_NAME = testConfig.TEST_SUITE_PREFIX + 'FirstName';
 const TEST_LAST_NAME = testConfig.TEST_SUITE_PREFIX + 'Surname';
-const EMAIL = testConfig.TEST_SUITE_PREFIX + 'email@test.com';
-const ROLE = 'Local Admin';
+const TEST_EMAIL = randomData.getRandomEmailAddress(TEST_WORKER_NUMBER);
+const TEST_ROLE = 'Internal - Administrator - Local';
+const TEST_INVALID_EMAIL = 'invalid_email@test.com';
+
 Scenario('I as an admin user should be able to modify a user account', async ({ I }) => {
-    // I.amOnPage('/');
-    // I.click('Continue');
     I.loginAsAdmin();
-    // I.amOnPage('/admin-dashboard');
+    I.waitForText('Your Dashboard');
+    I.createAdminAccount(TEST_FIRST_NAME, TEST_LAST_NAME, TEST_EMAIL, TEST_ROLE);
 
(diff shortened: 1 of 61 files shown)


src/test/end-to-end/tests/admin/user-management-test.ts
```diff
@@ -0,0 +1,84 @@
...
+    I.waitForText('User Updated');
+
+    I.click('Home');
+    I.deleteAccount(TEST_EMAIL);
+});
+
+Scenario('I as an admin user should be able to see all errors related to modify a user account', async ({ I }) => {
+    I.loginAsAdmin();
+    I.waitForText('Your Dashboard');
+    I.createAdminAccount(TEST_FIRST_NAME, TEST_LAST_NAME, TEST_EMAIL, TEST_ROLE);
```
KianKwa, line 64: What is the purpose of creating the account and deleting here? I can see this account getting used in this scenario

src/test/end-to-end/tests/admin/user-management-test.ts
```diff
@@ -0,0 +1,84 @@
...
+    I.click('Continue');
+    I.waitForText('Manage ' + TEST_EMAIL);
+    I.click('Change');
+    I.waitForText('What role would you like ' + TEST_EMAIL + ' to have?');
+    I.selectOption('updatedRole', 'CTSC Admin');
+    I.click('Continue');
+    I.waitForText('User Updated');
+
+    I.click('Home');
+    I.deleteAccount(TEST_EMAIL);
```
KianKwa, line 58: We also need to test the delete account functionality is working as intended. May be we should go back to the main  screen after account deletion and check that we get the error message "There is no user matching that email address" if we try to search for TEST_EMAIL

src/test/end-to-end/tests/admin/user-management-test.ts
NatashaAlker: The scenario is to alter the role for the account and I was asked not to perform this on an existing account :)

src/test/end-to-end/tests/admin/user-management-test.ts
NatashaAlker: I've added this check for completeness :)

KianKwa (dismissed):

src/test/end-to-end/tests/admin/user-management-test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+Feature('Manage admin user accounts');
+
+const TEST_WORKER_NUMBER = randomData.getRandomNumber(10000000, 99999999);
+const TEST_FIRST_NAME = testConfig.TEST_SUITE_PREFIX + 'FirstName';
+const TEST_LAST_NAME = testConfig.TEST_SUITE_PREFIX + 'Surname';
+const TEST_EMAIL = randomData.getRandomEmailAddress(TEST_WORKER_NUMBER);
+const TEST_ROLE = 'Internal - Administrator - Local';
+const TEST_INVALID_EMAIL = 'invalid_email@test.com';
+
```
ashwini-mv, line 12: All test email addresses should end up with either justice,gov.uk or hmcts.net, even the invalid addresses :)

ashwini-mv (changes requested):

KianKwa (approved):

ashwini-mv (approved):

PR #959, opened 2023-06-05
PUB-1945 Subscription by party name

KianKwa:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-1945

### Change description

Subscription by party name

image

image

image

image

image

```
[ ] Yes
[x] No
```


src/main/service/publicationService.ts
```diff
@@ -65,34 +87,124 @@ export class PublicationService {
...
+                artefact.search.parties.forEach(party => {
+                    const partyNames = this.constructPartyNames(party);
+                    party.cases.forEach(singleCase => {
+                        if (singleCase.caseName?.toLowerCase().includes(value.toLowerCase())) {
+                            this.storeUniquePartyCases(matches, singleCase, partyNames);
+                        }
+                    });
+                });
+            } else {
+                artefact.search.cases.forEach(singleCase => {
```
ChrisS1512, line 123: Just double checking if this else is needed. Does the parties object only display in the search result if parties are present? If it always displays I don't believe the else would ever be triggered

ChrisS1512 (changes requested):

src/main/service/publicationService.ts
KianKwa: For artefacts uploaded before this PR is merged, they will not have any "parties" object so if we search for case name, we will need to consider the "cases" object. I guess this can be removed eventually if we know that all artefacts contain the "parties" search value but we need that now.

ChrisS1512: Ah yes, good shout

ChrisS1512 (approved):

joshblackmoor (approved):

PR #1005, opened 2023-07-04
Pub 1963 - SJP postcode filters

NatashaAlker:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-1963

### Change description

Alter the behaviour of filters search to enable search by postal area (eg BD).

Auto select postcode items that fit search term.

```
[ ] Yes
[X] No
```


src/main/views/macros/sjp-common.njk
```diff
@@ -116,17 +116,77 @@
         // This function is used to dynamically show/hide individual filter checkbox items based on the search box value
         function searchFilterItems() {
             const filters = document.getElementsByClassName('govuk-checkboxes__item');
+            localStorage.setItem('searchFiltersValue', searchFilters.value);
+            var postcodeMatches = 0;
 
             for (var i = 0; i < filters.length; i++) {
```
ChrisS1512, line 122: Rather than using a for  loop here, would it be possible to use a for each loop instead?

By doing this it will save doing the [i] on each of the lines. Line 123 will have to change too, but I'm hoping it can be retrieved from the filters object.

I'm happy this being done in another ticket, as technically wasn't introduced as part of this. Either way is fine :)

ChrisS1512 (changes requested):

KianKwa (dismissed):

ChrisS1512 (dismissed):

KianKwa (approved):

junaidiqbalmoj (approved):

PR #1065, opened 2023-08-09
Pub 1651

NatashaAlker:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-1651

### Change description

List clean up items to standardise list presentation

```
[ ] Yes
[X] No
```


src/main/views/crown-daily-list.njk
```diff
@@ -13,26 +13,15 @@
 {% block content %}
   <div class="parent-box">
     <h2 id='page-heading' class="govuk-heading-l">
-      {{ courtName }}: {{ headerPart1 }}
+      {{ title }} for {{ courtName }}
       {% if bill %}
-        <br />/{{ courtName }}: {{ headerPart2 }}
```
KianKwa, line 18: If headerPart1 and headerPart2 not used anywhere else, they could be removed from the language json files

src/main/views/crown-daily-list.njk
```diff
@@ -13,26 +13,15 @@
...
+        <br />/{{ title }} for {{ courtName }}
```
ChrisS1512, line 18: The 'for' here should also be in welsh. I reckon the cleanest way is to do it similar to the primary health list: https://github.com/hmcts/pip-frontend/pull/1064/files#diff-6b14d1e62c549c362fa50bf678803f0885d920e16d927867271a431e9e14fcb7

to avoid having too many "/" in the bilingual versions

ChrisS1512 (changes requested):

KianKwa (dismissed):

KianKwa (approved):

ChrisS1512 (approved):

PR #1155, opened 2023-11-09
PUB-2206 - Dependency Updates

ChrisS1512:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-2206

### Change description

Dependency updates

```
[ ] Yes
[x] No
```


package.json
```diff
@@ -41,43 +41,43 @@
...
+        "@types/nunjucks": "^3.2.6",
+        "@types/require-directory": "^2.1.5",
+        "@types/serve-favicon": "^2.5.7",
+        "applicationinsights": "^2.9.0",
+        "awilix": "^9.0.0",
+        "axios": "^1.6.1",
         "axios-oauth-client": "^1.5.0",
         "axios-token-interceptor": "^0.2.0",
-        "codeceptjs": "^3.5.5",
+        "codeceptjs": "3.5.6",
```
KianKwa, line 63: Just checking the '^' symbol was removed intentionally to only accept the fixed version?

ChrisS1512: Yeah that's correct. 3.5.7 was causing extreme slowness in the tests alongside timeouts. Can't see anything in the changelog that could have caused it, so potentially a bug.

PR #1171, opened 2023-12-01
PUB-2112 Replace console.log with Logger

KianKwa:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-2112

### Change description

Replace console.log with Logger

```
[ ] Yes
[x] No
```


ChrisS1512 (changes requested):
There's one more example of console.log in the locationRequests file that needs replacing with the logger

ChrisS1512 (approved):

PR #1172, opened 2023-12-04
Pub 1977b - Add OPA Public List

NatashaAlker:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-1977

### Change description

Add OPA Public List style guide

```
[ ] Yes
[X] No
```


src/main/service/listManipulation/OpaPublicListService.ts
```diff
@@ -0,0 +1,114 @@
...
+        }
+        return null;
+    }
+
+    private buildIndividualName(individual) {
+        const firstName = ListParseHelperService.writeStringIfValid(individual?.individualFirstName);
+        const middleNames = ListParseHelperService.writeStringIfValid(individual?.individualMiddleName);
+        const surname = ListParseHelperService.writeStringIfValid(individual?.individualSurname);
+        const forenames = [firstName, middleNames].filter(n => n.length > 0).join(' ');
+        return [surname, forenames]
```
KianKwa, line 71: I am not sure if this is right for OPA public list. The example screenshots in the ticket has name in the following format "forename middleName surname" but you are returning "surname, firstname middleName". Has this been changed from the ticket?

src/main/service/listManipulation/OpaPublicListService.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    private buildOffenceDetails(offence) {
+        return {
+            offenceTitle: ListParseHelperService.writeStringIfValid(offence.offenceTitle),
+            offenceSection: ListParseHelperService.writeStringIfValid(offence.offenceSection),
+            offenceReportingRestriction: offence.reportingRestrictionDetail?.filter(n => n.length > 0).join(', '),
+        };
+    }
+
+    private buildCaseDetails(hearingCase) {
```
KianKwa, line 93: The logic here does not look correct. If there are multiple cases within the same hearing, currently you are looping through the case array and replacing the previous values with the last entry. You end up with just the record from the last case

src/main/resources/locales/en/opa-public-list.json
```diff
@@ -0,0 +1,24 @@
...
+    "publishedAt": "at",
+    "restrictionInformationHeading": "Restrictions on publishing or writing about these cases",
+    "restrictionInformationP1": "You must check if any reporting restrictions apply before publishing details on any of the cases listed here either in writing, in a broadcast or by internet, including social media.",
+    "restrictionInformationBoldText": "You'll be in contempt of court if you publish any information which is protected by a reporting restriction. You could get a fine, prison sentence or both.",
+    "restrictionInformationP2": "Specific restrictions ordered by the court will be mentioned on the cases listed here.",
+    "restrictionInformationP3": "However, restrictions are not always listed. Some apply automatically. For example, anonymity given to the victims of certain sexual offences.",
+    "restrictionInformationP4": "To find out which reporting restrictions apply on a specific case, contact:",
+    "restrictionBulletPoint1": "the court directly",
+    "restrictionBulletPoint2": "HM Courts and Tribunals Service on 0330 808 4407",
+    "restrictionInformationP5": "You can also read the",
```
KianKwa, line 14: You can remove the following lines. It has been confirmed with Chris Carter we dont need to display the reporting restriction link anymore because the link does not exists

restrictionInformationP5
restrictionLink

src/main/resources/locales/bill/opa-public-list.json
```diff
@@ -0,0 +1,32 @@
...
+    "publishedAt": "at/am",
+    "restrictionInformationHeading": "Restrictions on publishing or writing about these cases",
+    "restrictionInformationP1": "You must check if any reporting restrictions apply before publishing details on any of the cases listed here either in writing, in a broadcast or by internet, including social media.",
+    "restrictionInformationBoldText": "You'll be in contempt of court if you publish any information which is protected by a reporting restriction. You could get a fine, prison sentence or both.",
+    "restrictionInformationP2": "Specific restrictions ordered by the court will be mentioned on the cases listed here.",
+    "restrictionInformationP3": "However, restrictions are not always listed. Some apply automatically. For example, anonymity given to the victims of certain sexual offences.",
+    "restrictionInformationP4": "To find out which reporting restrictions apply on a specific case, contact:",
+    "restrictionBulletPoint1": "the court directly",
+    "restrictionBulletPoint2": "HM Courts and Tribunals Service on 0330 808 4407",
+    "restrictionInformationP5": "You can also read the",
```
KianKwa, line 14: The following 2 lines could be removed as per comment above

src/main/views/opa-public-list.njk
```diff
@@ -0,0 +1,139 @@
...
+                        </strong>
+                    </div>
+                    <p class="govuk-body">{{ restrictionInformationP2Cy }}</p>
+                    <p class="govuk-body">{{ restrictionInformationP3Cy }}</p>
+                    <p class="govuk-body">{{ restrictionInformationP4Cy }}</p>
+                    <ul class="govuk-list govuk-list--bullet">
+                        <li>{{ restrictionBulletPoint1Cy }}</li>
+                        <li>{{ restrictionBulletPoint2Cy }}</li>
+                    </ul>
+                    <p class="govuk-body">{{ restrictionInformationP5Cy }} <a href="#" class="govuk-link">{{ restrictionLinkCy }}</a> </p>
```
KianKwa, line 56: Don't need to display the link so this line can go

src/test/unit/mocks/opaPublicList.json
```diff
@@ -0,0 +1,113 @@
...
+                                                            "organisationName": "organisationName",
+                                                            "organisationAddress": {
+                                                                "line": [
+                                                                    "ORGANISATION ADDRESS"
+                                                                ],
+                                                                "town": "town name",
+                                                                "county": "",
+                                                                "postCode": ""
+                                                            },
+                                                            "offence": [
```
KianKwa, line 93: You can remove the the offence line from the test file if no offences for PROSECUTING_AUTHORITY

src/test/unit/service/listManipulation/opaPublicListService.test.ts
```diff
@@ -0,0 +1,48 @@
...
+import path from 'path';
+import { OpaPublicListService } from '../../../../main/service/listManipulation/OpaPublicListService';
+
+const opaPublicListService = new OpaPublicListService();
+
+describe('OPA Public List Service', () => {
+    const rawData = fs.readFileSync(path.resolve(__dirname, '../../mocks/opaPublicList.json'), 'utf-8');
+    const input = JSON.stringify(JSON.parse(rawData));
+
+    it('should format individual defendant name', async () => {
```
KianKwa, line 12: Might be worth test defendant from organisation details as I can see you have logic in the service class for that. Just need to add that case in the mock test file

src/main/service/listManipulation/OpaPublicListService.ts
```diff
@@ -0,0 +1,111 @@
...
+        }
+        return null;
+    }
+
+    private buildIndividualName(individual) {
+        const firstName = ListParseHelperService.writeStringIfValid(individual?.individualFirstName);
+        const middleNames = ListParseHelperService.writeStringIfValid(individual?.individualMiddleName);
+        const surname = ListParseHelperService.writeStringIfValid(individual?.individualSurname);
+        const forenames = [firstName, middleNames].filter(n => n.length > 0).join(' ');
+        return [forenames, surname]
```
KianKwa, line 73: This line here and the line above can be combined because the separators are the same. Also there is no need to use toString() at the end because they are all strings:

return [firstName, middleNames, surname].filter(n => n.length > 0).join(' ');

src/main/views/opa-public-list.njk
```diff
@@ -0,0 +1,138 @@
...
+                {% for row in listData %}
+                    {% set _ = opaCases.push([
+                        {
+                            text: row.name
+                        },
+                        {
+                            text: row.caseUrn
+                        },
+                        {
+                            text: row.offence[0].offenceTitle + " - " + row.offence[0].offenceSection + "\n Reporting Restriction - " + row.offence[0].offenceReportingRestriction
```
KianKwa, line 75: This doesn't look right. You only use the first array member. If there are multiple offences, the rest will not be displayed. Also I think the new line character will not work in nunjucks/HTML. You might want to try the <br> tag

KianKwa (approved):

ChrisS1512 (approved):

ChrisS1512 (approved):

PR #1187, opened 2024-01-03
PUB-2282 simplify logic to handle null values

NatashaAlker:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-2282

### Change description

Redirect to error page where there are null values in subscription data

```
[ ] Yes
[X] No
```


src/main/controllers/SubscriptionManagementController.ts
```diff
@@ -7,18 +7,23 @@ const subscriptionService = new SubscriptionService();
...
-                req.lng,
-                Object.keys(req.query)[0]
-            );
-            res.render('subscription-management', {
-                ...cloneDeep(req.i18n.getDataByLanguage(req.lng)['subscription-management']),
-                ...subscriptionData,
-            });
-        } else {
-            res.render('error', req.i18n.getDataByLanguage(req.lng).error);
+        if (!req.user) {
```
KianKwa, line 10: I am not sure if this is the right fix. From what I can see in the code, 'subscriptionData.caseTableData' and 'subscriptionData.locationTableData' can never be null. If there are no subscriptions, these 2 values are always empty array. 

I got a feeling the actual failing tests are these 2 below even though Jenkins is showing it occurs on a different file:

SubscriptionManagementController.test.ts -> 'should render error page if data is null'
BulkUnsubscribeController.test.ts -> 'should render error page if data is null'

In these tests the following stubs are incorrect as the table rows can never be null in the code. The mock request sometimes has the user defined/undefined causing the behaviour we are seeing.

sinon.stub(SubscriptionService.prototype, 'generateCaseTableRows').returns(null);
sinon.stub(SubscriptionService.prototype, 'generateLocationTableRows').returns(null);

PR #1194, opened 2024-01-17
Pub 2089 - Align JOH functionality

NatashaAlker:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-2089

### Change description

Update style guides of impacted lists to align the way judiciary is displayed

```
[ ] Yes
[X] No
```


src/main/service/listParseHelperService.ts
```diff
@@ -189,23 +189,28 @@ export class ListParseHelperService {
      */
     public findAndManipulateJudiciary(session: object): string {
         let judiciaries = '';
```
KianKwa, line 191: Just like the comment in channel-management. The logic can be simplified so we do not need to keep checking for non-empty string, add delimiter then trim it off later. You could do something like below:

```
let presidingJudiciary = '';
const judiciaries = [];
session['judiciary']?.forEach(judiciary => {
    if (judiciary?.isPresiding === true) {
        presidingJudiciary = ListParseHelperService.writeStringIfValid(judiciary?.johKnownAs);
    } else {
        judiciaries.push(ListParseHelperService.writeStringIfValid(judiciary?.johKnownAs));
    }
});
return [presidingJudiciary, ...judiciaries].filter(j =>  j.trim().length > 0).join(', ');
```

src/main/service/listParseHelperService.ts
```diff
@@ -188,24 +188,28 @@ export class ListParseHelperService {
...
-            if (judiciary?.isPresiding === true) {
-                judiciaries = ListParseHelperService.writeStringIfValid(judiciary?.johKnownAs);
-                foundPresiding = true;
-            } else if (!foundPresiding) {
-                if (ListParseHelperService.writeStringIfValid(judiciary?.johKnownAs) !== '') {
-                    judiciaries += ListParseHelperService.writeStringIfValid(judiciary?.johKnownAs) + ', ';
-                }
-            }
+            const name = ListParseHelperService.writeStringIfValid(judiciary?.johKnownAs);
+            judiciary?.isPresiding === true ? judiciaries.unshift(name) : judiciaries.push(name);
```
KianKwa, line 194: You are within the loop for each judiciary, so this judiciary definitely exists, so you can remove the ? operator after the judiciary in this line and the line above. Same with the crime method below.

src/main/service/listParseHelperService.ts
```diff
@@ -188,24 +188,28 @@ export class ListParseHelperService {
...
+     * Manipulate crime list judiciary data for writing out to screen.
+     * @param session
+     */
+    public findAndManipulateJudiciaryForCrime(session: object): string {
+        const judiciaries = [];
+        session['judiciary']?.forEach(judiciary => {
+            const name =
+                ListParseHelperService.writeStringIfValid(judiciary?.johTitle) +
+                ' ' +
+                ListParseHelperService.writeStringIfValid(judiciary?.johNameSurname);
```
KianKwa, line 209: The title could be blank and you don't want to add that extra space in that case. Better to do it like below when one of the fields is blank, it is not adding the space character:

const title = ListParseHelperService.writeStringIfValid(judiciary.johTitle);
const surname = ListParseHelperService.writeStringIfValid(judiciary.johNameSurname);
[title, surname].filter(j => j.length > 0).join(' ');

KianKwa (approved):

ChrisS1512 (approved):

PR #1197, opened 2024-01-26
PUB-2291- Updated OPA Public and Press Lists

MindGames2015:

-   [x] commit messages are meaningful and follow good commit message guidelines
-   [ ] README and other documentation has been updated / added (if needed)
-   [X] tests have been updated / new tests has been added (if needed)

### JIRA link

https://tools.hmcts.net/jira/browse/PUB-2291

### Change description

```
[ ] Yes
[X] No
```


src/main/service/listManipulation/OpaPublicListService.ts
```diff
@@ -5,22 +5,25 @@ import { OpaPressListService } from './OpaPressListService';
...
                         sitting.hearing.forEach(hearing => {
-                            const allDefendants = this.processPartyRoles(hearing);
-                            allDefendants.forEach(defendant => {
-                                hearing.case.forEach(hearingCase => {
-                                    const caseDetails = this.buildCaseDetails(hearingCase);
-                                    rows.push({
-                                        ...caseDetails,
-                                        ...defendant,
+                            hearing.case.forEach(hearingCase => {
+                                if (hearingCase['party']) {
```
KianKwa, line 17: Do not need the if statement here. Within the method processPartyRoles(), it is using "hearingCase.party?.forEach(...)". The question mark after the party is already checking whether party exist or not

src/test/unit/mocks/opaPublicList.json
```diff
@@ -44,114 +140,210 @@
                                                     },
                                                     {
                                                         "caseUrn": "URN5678",
+                                                        "party": [
```
KianKwa, line 143: This block of party information is identical to the previous block in the same case. The previous case already tested multiple defendants, you probably just need one defendant here rather than repeating the same test case.

src/main/service/listManipulation/OpaPublicListService.ts
MindGames2015: Removed if statement. Thanks Kian

KianKwa (dismissed):

src/main/service/listManipulation/OpaPressListService.ts
```diff
@@ -13,32 +13,33 @@ export class OpaPressListService {
...
-                                    const row = { ...caseInfo, ...defendant };
-                                    rows.push(row);
+                            hearing.case.forEach(hearingCase => {
+                                const defendantInfo = this.processPartyRoles(hearingCase);
+
+                                // Each case can have multiple defendants. They will be shown as separate entry on the list
+                                // as each can have its own plea date
+                                defendantInfo.forEach(defendant => {
+                                    const rows = [];
+                                    hearing.case.forEach(hearingCase => {
```
ChrisS1512, line 23: Found a bug here during testing. It seems to be duplicating parties, and I believe this is down to this line.

At line 16, it is looping through each case, therefore line 23 is no longer needed as it is looping through the cases a second time. This is then causing a cartesian product multiplication, where parties and cases are multiplied together.

A similar thing appears to be happening in channel management too, however I have not dug into the exact reason why (I imaging similar as it is causing a similar result)

ChrisS1512 (changes requested):

src/main/service/listManipulation/OpaPressListService.ts
MindGames2015: Modified OpaPressListService.ts, and ran OpaPressListService.ts test, which was successful.

PR #1218, opened 2024-02-28
PUB-2179 - Remove non-functional scrollbars

NatashaAlker:

### JIRA link

https://tools.hmcts.net/jira/browse/PUB-2179

### Change description

Remove all vertical scrollbars from list displays.
Remove non-functional horizontal scrollbars.

```
[ ] Yes
[X] No
```


ChrisS1512 (changes requested):
Theres one more list type that this needs sorting on which is the Primary Health List. 

Once that's done happy to approve

ChrisS1512 (approved):

src/main/views/care-standards-list.njk
```diff
@@ -31,7 +31,7 @@
 
     {{ searchInput(text = searchCases) }}
     <div class="parent-box overflow-table govuk-!-margin-top-5 search-area">
-      <table class="govuk-table" data-module="moj-sortable-table">
+      <table class="govuk-table" data-module="moj-sortable-table overflow-table">
```
ChrisS1512, line 34: I've found a bug which is causing the sorting on the table to no longer work. I believe this is due to the overflow-table property being against the data-module - I believe this needs to be under class.

ChrisS1512 (changes requested):

KianKwa (approved):

PR #1237, opened 2024-03-22
PUB-2306 Family and mixed lists - Move party to case level

KianKwa:

https://tools.hmcts.net/jira/browse/PUB-2306

### Change description ###

Family and mixed lists - Move party to case level

```
[ ] Yes
[x] No
```


src/main/service/listManipulation/CivilFamilyAndMixedListService.ts
```diff
@@ -15,11 +15,39 @@ export class CivilFamilyAndMixedListService {
      * Manipulate the family/mixed json data for rendering to the screen.
      * @param list
      */
-    public sculptedFamilyMixedListData(list: string): object {
+    public sculptedFamilyMixedListData(list: string, partyAtHearingLevel = false): object {
```
ChrisS1512, line 18: I'm wondering if this method adds value. A similar if statement is done in the controller. Could the controller just call the sculpted list data methods this one calls directly?

src/main/service/listManipulation/CivilFamilyAndMixedListService.ts
```diff
@@ -15,11 +15,39 @@ export class CivilFamilyAndMixedListService {
...
+        if (partyAtHearingLevel) {
+            return this.sculptedListDataV1(list, true);
+        }
         return this.sculptedListData(list, true);
     }
 
     private sculptedListData(list: string, isFamilyMixedList = false): object {
```
ChrisS1512, line 25: I think this method / equivilant V1 below would be better replaced with something like sculptedListDataHearingLevel etc, just to make it clear. What do you think?

ChrisS1512 (changes requested):

src/main/service/listManipulation/CivilFamilyAndMixedListService.ts
KianKwa: Renamed to sculptedListDataPartyAtHearingLevel()

src/main/service/listManipulation/CivilFamilyAndMixedListService.ts
KianKwa: Have removed the top level methods so the controller calls them directly.

ChrisS1512 (dismissed):

ChrisS1512 (approved):

PR #1250, opened 2024-04-05
PUB-2029

alexbottenberg:
https://tools.hmcts.net/jira/browse/PUB-2029

Adding media request notification banner to admin dashboard

```
[ ] Yes
[x] No
```


MindGames2015 (approved):

src/main/views/admin-dashboard.njk
```diff
@@ -13,6 +14,25 @@
     <div class="govuk-grid-row parent-box">
         <h1 class="govuk-heading-l">{{ heading }}</h1>
     </div>
+    {% set html %}
```
KianKwa, line 17: A small issue. It might be better to move the "set" to after line 31 to this does not need to be set for other user roles as it is unnecessary.

KianKwa (approved):

PR #1272, opened 2024-05-09
PUB-2352 Create third party user on system admin dashboard

KianKwa:
### JIRA link

[https://tools.hmcts.net/jira/browse/<PR-name-here>](https://tools.hmcts.net/jira/browse/PUB-2352)

### Change description

Create third party user on system admin dashboard

```
[ ] Yes
[x] No
```


src/main/controllers/CreateThirdPartyUserSuccessController.ts
```diff
@@ -0,0 +1,22 @@
...
+        const formData = req.cookies?.formCookie ? JSON.parse(req.cookies['formCookie']) : {};
+        res.render('create-third-party-user-success', {
+            ...cloneDeep(req.i18n.getDataByLanguage(req.lng)['create-third-party-user-success']),
+            formData,
+        });
+
+        CreateThirdPartyUserSuccessController.removeThirdPartyUserFromFormCookie(res, formData);
+    }
+
+    private static removeThirdPartyUserFromFormCookie(res, formData) {
```
ChrisS1512, line 16: Rather than specifically removing the values, could we clear the cookie instead?

src/main/controllers/CreateThirdPartyUserController.ts
```diff
@@ -0,0 +1,36 @@
...
+        if (formErrors) {
+            res.render('create-third-party-user', {
+                ...cloneDeep(req.i18n.getDataByLanguage(req.lng)['create-third-party-user']),
+                userRoleList: thirdPartyService.buildThirdPartyRoleList(formData.thirdPartyRole),
+                formData,
+                formErrors,
+            });
+        } else {
+            formData.thirdPartyRoleObject = thirdPartyService.getThirdPartyRoleByKey(formData.thirdPartyRole);
+            res.cookie('formCookie', JSON.stringify(formData), { secure: true });
```
ChrisS1512, line 32: Is the expiry of this just the current session? I'm not sure what the default is, but it would be good for it to not last indefinitely, preferably deleted when the user closes tab / browser

src/main/resources/requests/AccountManagementRequests.ts
```diff
@@ -33,17 +33,17 @@ export class AccountManagementRequests {
...
-    public async createPIAccount(payload, requester): Promise<boolean> {
+    public async createPIAccount(payload, requester): Promise<object | null> {
         try {
-            const response = await accountManagementApi.post('/account/add/pi', payload, {
+            const response = await accountManagementApi.post('/account/v2/pi', payload, {
                 headers: { 'x-issuer-id': requester },
             });
             logger.info('P&I account created');
-            return response.status === StatusCodes.CREATED;
+            return response.status === StatusCodes.CREATED ? response.data : null;
```
ChrisS1512, line 42: Is the response body required here in the return? I had a look through and can only see it being used for a true / false check. If it's not being used, I think the existing boolean return here is clearer at showing the success / failure

src/main/controllers/CreateThirdPartyUserSuccessController.ts
KianKwa: The formCookie is used for other things too in the application. I was just being a bit cautious ant not wanting to delete other information stored there for other pages that might still be required when we revisit those pages. Can clear the cookie here if u think that's ok

src/main/resources/requests/AccountManagementRequests.ts
KianKwa: There problem with the true/false check is the backend endpoint is returning a 201 success status too if the endpoint returns ERRORED_ACCOUNTS rather than CREATED_ACCOUNTS, so there is no way to differentiate whether the account is successfully created or not

src/main/controllers/CreateThirdPartyUserSuccessController.ts
KianKwa: I have now cleared the cookie after third party user successfully created

src/main/controllers/CreateThirdPartyUserSuccessController.ts
```diff
@@ -10,13 +10,6 @@ export default class CreateThirdPartyUserSuccessController {
...
 
-        CreateThirdPartyUserSuccessController.removeThirdPartyUserFromFormCookie(res, formData);
-    }
-
-    private static removeThirdPartyUserFromFormCookie(res, formData) {
-        formData.thirdPartyName = '';
-        formData.thirdPartyRole = '';
-        formData.thirdPartyRoleObject = null;
-        res.cookie('formCookie', JSON.stringify(formData), { secure: true });
+        res.cookie('formCookie', '', { secure: true });
```
ChrisS1512, line 13: I think the clearCookie command can be used rather than setting it to blank: https://expressjs.com/en/api.html#res.clearCookie

ChrisS1512 (changes requested):

ChrisS1512 (approved):

PR #1280, opened 2024-05-21
Pub 2361

alexbottenberg:
https://tools.hmcts.net/jira/browse/PUB-2361

Adding multi-select checkboxes for list removal 

```
[ ] Yes
[x] No
```


src/main/controllers/RemoveListSearchResultsController.ts
```diff
@@ -6,20 +6,35 @@ import { SummaryOfPublicationsService } from '../service/SummaryOfPublicationsSe
...
+                      await summaryOfPublicationsService.getPublications(locationId, req.user?.['userId'], true)
                   ),
+                  noOptionSelectedError: noOptionSelectedError,
               })
             : res.render('error', req.i18n.getDataByLanguage(req.lng).error);
     }
+
+    public async post(req: PipRequest, res: Response): Promise<void> {
+        if (req.user) {
+            if (req.body.courtLists == undefined) {
```
ChrisS1512, line 30: I would add a '?' after the body. Very unlikely for it to be sent with no body, but I think it's worth capturing:

'req.body?.courtLists'

I'm also not sure you need the == undefined - I think 

if (req.body?.courtLists) 

will work

src/main/views/remove-list-search-results.njk
```diff
@@ -12,39 +13,59 @@
...
+                    {% for item in removalList %}
+                        {% set listCount = listCount + 1 %}
+                        <tr class="govuk-table__row">
+                            <td class="govuk-table__cell">{{ item.listTypeName }}</td>
+                            <td class="govuk-table__cell">{{ court.name }}</td>
+                            <td class="govuk-table__cell" data-sort-value="{{ item.contDate | dateWithShortMonthNameToSortValue }}">{{ item.contDate }}</td>
+                            <td class="govuk-table__cell">{{ item.dateRange }}</td>
+                            <td class="govuk-table__cell">{{ item.language | titleCase }}</td>
+                            <td class="govuk-table__cell">{{ item.sensitivity | titleCase }}</td>
+                            <td class="govuk-table__cell govuk-table__cell--numeric">
```
ChrisS1512, line 54: The checkbox I think should be centre'd if possible under the select header. It's currently left aligned

image

src/main/controllers/RemoveListConfirmationController.ts
```diff
@@ -14,44 +14,37 @@ const userManagementService = new UserManagementService();
...
+        res.render('remove-list-confirmation', {
+            ...cloneDeep(req.i18n.getDataByLanguage(req.lng)['remove-list-confirmation']),
+            removalList: manualUploadService.formatListRemovalValues(listData),
+            locationId: formData.locationId,
+            court: await courtService.getLocationById(formData.locationId),
+            displayError: false,
+        });
     }
 
     public async post(req: PipRequest, res: Response): Promise<void> {
```
ChrisS1512, line 32: There should be a check in this method to make sure req.body, req.body.artefactIds and req.body.locationId's exists.

Unlikely scenario but I think it will cause a crash if they're not there

src/main/controllers/RemoveListConfirmationController.ts
```diff
@@ -60,17 +53,64 @@ export default class RemoveListConfirmationController {
...
                     ...cloneDeep(req.i18n.getDataByLanguage(req.lng)['remove-list-confirmation']),
-                    artefact,
-                    court: await courtService.getLocationById(formData.locationId),
+                    court: await courtService.getLocationById(locationId),
+                    removalList: manualUploadService.formatListRemovalValues(listData),
                     displayError: true,
                 });
         }
     }
+
```
ChrisS1512, line 76: Can we move the below methods into a service class instead? It will keep the controller / service layer split clear + also slightly easier to test

ChrisS1512 (changes requested):

ChrisS1512 (approved):

ChrisS1512 (changes requested):
One of the functional tests is failing, I think it's due to the change in how lists are removed

ashwini-mv:
> One of the functional tests is failing, I think it's due to the change in how lists are removed

Fixed:)

ChrisS1512 (approved):

PR #1282, opened 2024-05-22
Pub 2453

alexbottenberg:
https://tools.hmcts.net/jira/browse/PUB-2453

Re-wording of the open justice statement in the IAC list

```
[ ] Yes
[x] No
```


KianKwa (changes requested):
The changes in the frontend is fine but we will need to update the PDF too to be consistent with the style guide

KianKwa (approved):

PR #1295, opened 2024-06-10
PUB-2442 SSO login prototype

KianKwa:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-2442

### Change description

SSO login prototype

```
[ ] Yes
[x] No
```


src/main/authentication/authentication.ts
```diff
@@ -23,10 +25,23 @@ async function piAadVerifyFunction(iss, sub, profile, accessToken, refreshToken,
...
-async function serializeUser(foundUser, done) {
-    if (foundUser['flow'] === 'CFT') {
-        const user = await accountManagementRequests.getPiUserByCftID(foundUser['uid']);
+async function ssoVerifyFunction(iss, sub, profile, accessToken, refreshToken, done): Promise<any> {
+    const userRole = await ssoAuthentication.determineUserRole(profile.oid, accessToken);
+    if (userRole) {
+        profile['roles'] = userRole;
+        profile['email'] = profile._json['preferred_username'];
+        profile['flow'] = 'SSO';
+        const response = await ssoAuthentication.handleSsoUser(profile);
```
ChrisS1512, line 34: For CFT IDAM. the creation is done in the 'SerializeUser' function instead so I think it makes sense to keep them consistent

src/main/authentication/ssoAuthentication.ts
```diff
@@ -0,0 +1,82 @@
...
+            [process.env.SSO_SG_SUPER_ADMIN_CTSC, 'INTERNAL_SUPER_ADMIN_CTSC'],
+            [process.env.SSO_SG_SUPER_ADMIN_LOCAL, 'INTERNAL_SUPER_ADMIN_LOCAL'],
+            [process.env.SSO_SG_ADMIN_CTSC, 'INTERNAL_ADMIN_CTSC'],
+            [process.env.SSO_SG_ADMIN_LOCAL, 'INTERNAL_ADMIN_LOCAL'],
+        ]);
+        let userRole;
+
+        if (userGroupsObject?.value.length > 0) {
+            const userGroups = userGroupsObject.value;
+            securityGroupMap.forEach((value: string, key: string) => {
```
ChrisS1512, line 45: Is there a convenience function that we can use instead of this like the java matchAny function instead? of manually looping through

src/main/routes/routes.ts
```diff
@@ -678,6 +679,20 @@ export default function (app: Application): void {
...
 
+    // SSO Routes
+    if (process.env.ENABLE_SSO === 'true') {
+        app.get('/sso-login', regenerateSession, keepSessionLanguage, (req, res, next) =>
+            passport.authenticate('sso', { failureRedirect: '/' })(req, res, next)
+        );
+
+        app.post(
+            '/sso',
+            (req, res, next) => passport.authenticate('sso', { failureRedirect: '/' })(req, res, next),
```
ChrisS1512, line 690: I think regenerateSession needs to be here too

ChrisS1512 (changes requested):

src/main/routes/routes.ts
KianKwa: This is the redirect URL from the SSO app. regenerateSession is already in the login route 'sso-login'. I don't think it is needed here again.

src/main/authentication/ssoAuthentication.ts
KianKwa: I have updated the code slightly using 'find'. This is still not as simple as you would like. They are kind of 2 arrays here. One is all the groups the user belongs to, and the other is a list of CaTH security groups. After we find the security group we also need to get the equivalent role for that group.

src/main/authentication/authentication.ts
KianKwa: The user creation needs to be done here for SSO for error handling. If we fail to create the system admin or admin account (particularly the system admin as it can fail when hitting the max size), I am setting the req.user.created to false, so it can render the appropriate error page in processSsoSignIn().
It was originally done in serializeUser() but req.user cannot be set there so I moved it here.

src/main/routes/routes.ts
ChrisS1512: You're right, on checking again I don't think it needs to be here.

ChrisS1512 (approved):

ChrisS1512 (dismissed):

PR #1308, opened 2024-07-15
PUB-2520 Use SJP delta URL when applying filters

KianKwa:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-2520

### Change description

Use SJP delta URL when applying filters

```
[ ] Yes
[x] No
```


src/main/controllers/style-guide/SjpPressListController.ts
```diff
@@ -94,10 +94,16 @@ export default class SjpPressListController {
 
     public async filterValues(req: PipRequest, res: Response): Promise<void> {
         if (validate(req.query?.artefactId as string)) {
+            const sjpPressMetaData = await publicationService.getIndividualPublicationMetadata(
```
ChrisS1512, line 97: This method can technical return null if user doesn't have permission / not found which I think would then cause a crash on line 101

ChrisS1512: Same comment applies to the public list controller too

ChrisS1512 (changes requested):

src/main/controllers/style-guide/SjpPressListController.ts
KianKwa: Updated the code to handle null metadata

ChrisS1512 (approved):

ChrisS1512 (approved):

PR #1323, opened 2024-08-12
Modified test to create random test user accounts with test prefix

ashwini-mv:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-2546

### Change description
Modified test to create random test user accounts with test prefix

```
[ ] Yes
[x] No
```


KianKwa (approved):

package.json
```diff
@@ -63,6 +63,7 @@
         "connect-redis": "7.1.1",
         "core-js": "3.37.1",
         "cors": "^2.8.5",
+        "csv": "^6.3.10",
```
ChrisS1512, line 66: I think this should be in the devDependencies section rather than dependencies

src/test/end-to-end/tests/system-admin/bulk-create-media-accounts-test.ts
```diff
@@ -13,7 +36,8 @@ Scenario('I as a system admin should be able to bulk create media accounts', asy
     I.see(
         'Note the upload process has a maximum of 30 accounts created per run. Please ensure the file uploaded for processing has no more than 30 cases.'
     );
-    I.attachFile('#bulk-account-upload', './shared/mocks/bulkCreateUserValid.csv');
+    I.wait(10);
```
ChrisS1512, line 39: Are the waits still needed here? If so I wonder if we can do this with something like I wait for text, rather than waiting a specific time

ChrisS1512 (changes requested):

src/test/end-to-end/tests/system-admin/bulk-create-media-accounts-test.ts
ashwini-mv: No, just forgot to remove that.  I will remove it:)

package.json
ashwini-mv: ok, I will move it under dev dependencies

ChrisS1512 (approved):

PR #1328, opened 2024-08-15
PUB-2327 Express rate limit prototype

KianKwa:
### JIRA link

https://tools.hmcts.net/jira/browse/<PR-name-here>

### Change description

Express rate limit prototype

```
[ ] Yes
[x] No
```


ChrisS1512 (changes requested):
Prototype looks great and what we need.

If not done already, I think it'd be worth testing this on one of the lower environments to make sure it works when deployed into AKS. Are you OK doing that as part of this ticket?

PR #1344, opened 2024-09-09
PUB-2548 - Updated typescript eslint

ChrisS1512:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-2548

### Change description

Upgraded to v8 of eslint

This involved changing a number of require statements to import. There were a few places were this was not possible due to requiring a top level await, which is only possible when the project is set as a module.

Also moved language files back to single folder due to not being supported via types

```
[ ] Yes
[x] No
```


src/main/development.ts
```diff
@@ -1,6 +1,7 @@
+/* eslint-disable @typescript-eslint/no-require-imports */
 import * as express from 'express';
 
-const setupDev = (app: express.Express, developmentMode: boolean): void => {
+export const setupDev = (app: express.Express, developmentMode: boolean): void => {
```
KianKwa, line 4: Do we still need module.exports at the bottom now we export the const here?

src/main/cacheManager.ts
```diff
@@ -21,7 +22,7 @@ export function setRedisCredentials(): any {
 
 const logger = Logger.getLogger('app');
 
-let redisClient;
+export let redisClient;
```
KianKwa, line 25: module.exports below could be removed too since it is aready marked as export here

src/main/modules/nunjucks/index.ts
```diff
@@ -29,15 +29,16 @@ export class Nunjucks {
                 express: app,
             }
         );
-        const addFilters = require('./njkFilters');
-        addFilters(env);
+        /* eslint-disable-next-line @typescript-eslint/no-require-imports */
+        const njkFilters = require('./njkFilters');
```
KianKwa, line 33: Are we not able to do the following?

import { createFilters } from './njkFilters';
createFilters(env);

ChrisS1512: Unfortunately not :( 

That kind of import needs to be at the top level and not within a a method.

You can have an import like this at method level:

const { createFilters} = await import './njkFilters', 

however due to the await it then needs an async. However unfortunately with this file being called from the app.ts, you cannot add an async unless you turn the application into a module

src/main/development.ts
ChrisS1512: Good spot, updated

src/main/cacheManager.ts
ChrisS1512: Good spot, updated

KianKwa (approved):

NatashaAlker (approved):

PR #1360, opened 2024-10-08
PUB-2608 add language preference

junaidiqbalmoj:
### JIRA link

https://tools.hmcts.net/jira/browse/PR-2588
https://tools.hmcts.net/jira/browse/PR-2589
https://tools.hmcts.net/jira/browse/PR-2608

### Change description

This PR contains all changes for location list type subscription. It contains changes for PUB-2588, PUB-2589 and PUB-2608.

**Add Subscription:**

1. Add list type selection page when court subscription is selected
2. Add list language page

**Configure List Type:**
Add functionality to configure list language

[ ] Yes
[x] No


src/main/controllers/SubscriptionConfigureListController.ts
```diff
@@ -2,30 +2,45 @@ import { Response } from 'express';
...
         });
     }
 
-    public async filterValues(req: PipRequest, res: Response): Promise<void> {
-        const filterValues = filterService.generateFilterKeyValues(req.body);
-        res.redirect(`subscription-configure-list?filterValues=${filterValues}`);
+    public async post(req: PipRequest, res: Response): Promise<void> {
+        const result = await subscriptionService.createListTypeSubscriptionPayload(req.body['list-selections[]']);
+
+        if (Object.values(result).length == 0) {
```
KianKwa, line 27: We should change the method "subscriptionService.createListTypeSubscriptionPayload()" to return string array and not object so you don't to use Object.values(result) conversion here. After that can just do "if (result.length)"

src/main/service/SubscriptionService.ts
```diff
@@ -493,11 +508,7 @@ export class SubscriptionService {
                 listType.jurisdictions.some(jurisdiction => courtJurisdictions.includes(jurisdiction)) &&
                 (listType.restrictedProvenances.length === 0 || listType.restrictedProvenances.includes(userRole))
             ) {
-                if (
-                    selectedListTypes == null ||
-                    selectedListTypes.length == 0 ||
-                    selectedListTypes.includes(listName)
-                ) {
+                if (selectedListTypes != null && selectedListTypes.includes(listName)) {
```
KianKwa, line 511: Do not need an if/else here. Can just do: 

listType.checked  = selectedListTypes != null && selectedListTypes.includes(listName);

src/main/views/subscription-config-list-language.njk
```diff
@@ -0,0 +1,83 @@
...
+                titleText: error.title,
+                errorList: [
+                    {
+                        text: error.noSelection,
+                        href: "#english"
+                    }
+                ]
+            }) }}
+        {% endif %}
+        <form action="subscription-configure-list-confirmed" method="post">
```
KianKwa, line 32: We nornally post back to the same page to check for error and not the confirmed page. Might be better to add a post method to subscription-config-list-language page for that

src/main/views/subscription-config-list-language.njk
```diff
@@ -0,0 +1,83 @@
...
+            <div class="govuk-form-group">
+                <fieldset class="govuk-fieldset" id="language-selector">
+                    <legend class="govuk-fieldset__legend govuk-fieldset__legend--l">
+                        <h1 class="govuk-fieldset__heading">
+                            {{ title }}
+                        </h1>
+                    </legend>
+                    <div class="govuk-radios" data-module="govuk-radios">
```
KianKwa, line 40: If that not possible to use the nunjucks version of govukRadios rather that raw HTML here?

src/main/controllers/SubscriptionConfigListLanguageController.ts
```diff
@@ -0,0 +1,27 @@
+import { PipRequest } from '../models/request/PipRequest';
```
KianKwa, line 1: We should rename this file and all related view and language files to SubscriptionConfigureLIstLanguage and not SubscriptionConfigLIstLanguage to be consistent with other files

src/main/controllers/SubscriptionConfigListLanguageController.ts
```diff
@@ -0,0 +1,27 @@
...
+import { SubscriptionService } from '../service/SubscriptionService';
+
+const subscriptionsService = new SubscriptionService();
+
+export default class SubscriptionConfigListLanguageController {
+    public async get(req: PipRequest, res: Response): Promise<void> {
+        const listTypeLanguageSubscriptions = await subscriptionsService.getUserSubscriptionListLanguage(
+            req.user['userId']
+        );
+        if (req.query.error === 'true') {
```
KianKwa, line 13: This feels wrong to check for query in the 'GET' method. We should add a POST method too for this class

src/main/controllers/SubscriptionConfigureListConfirmedController.ts
```diff
@@ -2,18 +2,39 @@ import { Response } from 'express';
...
-        const result = await subscriptionService.configureListTypeForLocationSubscriptions(
-            req.user['userId'],
-            req.body['list-selections[]'],
-            []
-        );
+        const userId = req.user['userId'];
+        const cacheService = new PendingSubscriptionsFromCache();
 
-        if (result) {
+        if (req.body['list-language'] === undefined) {
```
KianKwa, line 14: The check for language stuffs should go into the POST method in the list language controller. Need to update the nunjucks file to post to that route and not to this one.

src/main/controllers/SubscriptionConfigureListController.ts
junaidiqbalmoj: fixed.

src/main/service/SubscriptionService.ts
junaidiqbalmoj: fixed.

src/main/views/subscription-config-list-language.njk
junaidiqbalmoj: fixed.

src/main/views/subscription-config-list-language.njk
junaidiqbalmoj: fixed.

src/main/controllers/SubscriptionConfigListLanguageController.ts
junaidiqbalmoj: fixed.

src/main/controllers/SubscriptionConfigListLanguageController.ts
junaidiqbalmoj: fixed.

src/main/controllers/SubscriptionConfigureListConfirmedController.ts
junaidiqbalmoj: fixed.

KianKwa (dismissed):

KianKwa (approved):

PR #1364, opened 2024-10-10
PUB-2616 Update pages with A-Z

KianKwa:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-2616

### Change description

Update pages with A-Z

```
[ ] Yes
[x] No
```


src/main/resources/locales/cy/subscription-configure-list.json
NatashaAlker, line 1: I can see this has replaced some english text, but should it be in welsh?

KianKwa: ariaLabelNav is removed from both English and Welsh language resource file. The Welsh file text is never translated before but doesn't matter now as it is getting removed

NatashaAlker: My mistake! I got mixed up with the lines added above, ignore me :)

NatashaAlker (approved):

PR #1378, opened 2024-10-31
PUB-2647 - SSO Rejected Login

ChrisS1512:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-2647

### Change description

Added SSO Rejected Login page

```
[ ] Yes
[x] No
```


src/main/controllers/SsoRejectedLoginController.ts
```diff
@@ -0,0 +1,17 @@
...
+import { Response } from 'express';
+import { ssoNotAuthorised } from '../helpers/consts';
+
+/**
+ * Controller for handling SSO Rejected logins
+ */
+export default class SsoRejectedLoginController {
+    public async get(req: PipRequest, res: Response): Promise<void> {
+        const sessionMessages = req['session']?.messages;
+        if (Array.isArray(sessionMessages) && sessionMessages.length > 0 && sessionMessages[0] === ssoNotAuthorised) {
```
KianKwa, line 11: Does it needs to be the first array member of messages for it to be accepted. If not can we use the following which can work if messages are either for array or string?

if (sessionMessages && sessionMesssages.includes(ssoNotAuthorised))

KianKwa (approved):

PR #1459, opened 2025-04-04
PUB-2781 - Performance Tests

ChrisS1512:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-2781

### Change description

Added performance tests for Verified login

```
[ ] Yes
[x] No
```


src/test/end-to-end/tests/performance/verified-login-subscription-test.ts
```diff
@@ -0,0 +1,16 @@
+Feature('Login - Performance');
+
+//This value can be changed to determine how many tests should run.
+//The number of concurrent logins can be changed via the 'run-workers' setting, in the package.json
+for (let i = 0; i < 10; i++) {
```
KianKwa, line 5: Is this value 10 likely to be updated for performance run, or we mainly use the run-workers setting? If it needs to be changed frequently, are we better to use an environment variable?

KianKwa (approved):

PR #1466, opened 2025-04-22
PUB-2797 Add Backend Security Checks

junaidiqbalmoj:

https://tools.hmcts.net/jira/browse/PUB-2797

### Change description ###

Add Backend Security Checks

[ ] Yes
[x] No

### DIFF
--- a/src/main/resources/requests/PublicationRequests.ts
+++ b/src/main/resources/requests/PublicationRequests.ts
@@ -6,11 +6,11 @@ import { LogHelper } from '../logging/logHelper';
 const logHelper = new LogHelper();
 
 export class PublicationRequests {
-    public async getIndividualPublicationMetadata(artefactId, userId, admin): Promise<any> {
+    public async getIndividualPublicationMetadata(artefactId, requesterId, admin): Promise<any> {
         try {
             let header;
-            if (userId) {
-                header = { headers: { 'x-user-id': userId, 'x-admin': admin } };
+            if (requesterId) {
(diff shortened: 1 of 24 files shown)


src/main/resources/requests/PublicationRequests.ts
```diff
@@ -6,11 +6,11 @@ import { LogHelper } from '../logging/logHelper';
 const logHelper = new LogHelper();
 
 export class PublicationRequests {
-    public async getIndividualPublicationMetadata(artefactId, userId, admin): Promise<any> {
+    public async getIndividualPublicationMetadata(artefactId, requesterId, admin): Promise<any> {
         try {
             let header;
-            if (userId) {
```
ChrisS1512, line 12: Can we temporarily also have the x-user-id here. I'm thinking when we merge, it'd be good to merge frontend first, then merge backend and then after we can remove the old x-user-id

ChrisS1512: This is the same for anywhere where we have replaced the existing ID header that is passed with the new one.  Any ones where we are adding a new header where one didn't exist previously should be fine

ChrisS1512 (changes requested):

src/main/resources/requests/PublicationRequests.ts
junaidiqbalmoj: added x-user-id.

KianKwa (approved):

PR #1505, opened 2025-07-02
PUB-2905 Stop deletion of list types if there are remaining location subscription

KianKwa:

https://tools.hmcts.net/jira/browse/PUB-2905

### Change description ###

Stop deletion of list types if there are remaining location subscription

```
[ ] Yes
[x] No
```


src/main/service/SubscriptionService.ts
```diff
@@ -141,12 +141,46 @@ export class SubscriptionService {
...
+        }
+        return false;
     }
 
-    public async bulkDeleteSubscriptions(subscriptionIds: string[], userId: string): Promise<object> {
-        return subscriptionRequests.bulkDeleteSubscriptions(subscriptionIds, userId);
+    public async bulkDeleteSubscriptions(
+        subscriptionIds: string[],
+        userId: string,
+        userRole: string
```
ChrisS1512, line 155: I think this is userProvenance instead of role

ChrisS1512 (changes requested):

src/main/service/SubscriptionService.ts
KianKwa: I have changed them all to userProvenance

ChrisS1512 (approved):

ChrisS1512 (approved):

PR #1539, opened 2025-09-16
PUB-2852 PDDA Crown Daily List

KianKwa:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-2851

### Change description

PDDA Crown Daily List

```
[ ] Yes
[x] No
```


src/main/service/listManipulation/CrownPddaListService.ts
```diff
@@ -3,43 +3,54 @@ import { formatDate } from '../../helpers/dateTimeHelper';
...
             });
         });
         return sittings;
     }
 
-    private buildHearings(sitting: any): any[] {
+    private buildHearings(sitting: any, isDailyList: boolean): any[] {
         const hearings = [];
         sitting.Hearings.forEach(hearing => {
+            let representativeName;
```
ChrisS1512, line 42: This can be simplified here by assigning the default on line 42, and then the first part of the if statement isn't needed

PR #1552, opened 2025-09-30
PUB-3003 Updated getPublicationsByLocation calls on frontend

KianKwa:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-3003

### Change description

Updated getPublicationsByLocation calls on frontend

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/controllers/style-guide/CrownPddaListController.ts
+++ b/src/main/controllers/style-guide/CrownPddaListController.ts
@@ -1,0 +1,65 @@
+import { PipRequest } from '../../models/request/PipRequest';
+import { Response } from 'express';
+import { formatMetaDataListType, isUnexpectedListType, isValidList, isValidListType } from '../../helpers/listHelper';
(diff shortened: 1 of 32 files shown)


src/main/controllers/SummaryOfPublicationsController.ts
```diff
@@ -26,10 +24,7 @@ export default class SummaryOfPublicationsController {
                     req.lng === 'cy' ? locationMetadata.welshCautionMessage : locationMetadata.cautionMessage;
             }
 
-            const publications = await summaryOfPublicationsService.getPublications(
```
ChrisS1512, line 29: I think we should keep the the parseInt in some way here, as it adds a level of validation on the input data via the query param. Similar comment on the RemoveListSearchResultsController below.

Potentially better as an if statement that throws an error if not an int?

ChrisS1512 (changes requested):

PR #1676, opened 2026-04-23
PUB-2902 - Hide hearing lists

NatashaAlker:
### JIRA link

https://tools.hmcts.net/jira/browse/PUB-2902

### Change description

Hide SSCS Daily List and SSCS Daily List - Additional Hearings lists as they are not yet ready for publishing and may confuse users during the publishing / subscription process.

```
[ ] Yes
[x] No
```


src/main/helpers/consts.ts
```diff
@@ -62,3 +62,5 @@ export const reSignInUrls = {
 };
 
 export const ssoNotAuthorised = 'NOT_AUTHORISED';
+
+export const hiddenListTypes = new Set(['SSCS_DAILY_LIST', 'SSCS_DAILY_LIST_ADDITIONAL_HEARINGS']);
```
KianKwa: The available list types all come from the same list type lookup file. I think it is better to modify that than creating another new hidden list to check. If they need to be added back a year later and something else is working on this it is not clear to them we have a bespoke set to hide the list types. It is clearer if the list type management are all in the same place.

The list types are all hardcoded so they can simply be removed from the lookup file and re-added if needed. Alternatively create a new flag for hidden list types there and check against the new flag.

KianKwa (approved):

PR #1677, opened 2026-04-24
Pub-3229: Adding hints to subscription-add radio buttons

alexbottenberg:
https://tools.hmcts.net/jira/browse/PUB-3229

Adding hints to subscription-add radio buttons

```
[ ] Yes
[x] No
```


src/test/unit/views/subscription-add.test.ts
```diff
@@ -96,6 +98,14 @@ describe('Subscriptions add Page initial load', () => {
             'Could not find the radio button with label ' + expectedRadioLabel3
         );
     });
+
+    it('should display radio button hint', () => {
```
junaidiqbalmoj: I can see that you added test for one radio button but missing test for other radio button. can you add another test for the missing one as well?

junaidiqbalmoj (changes requested):

junaidiqbalmoj (approved):

PR #1710, opened 2026-07-28
Add publishing policy page

RebeccaHayleyPickles:
https://tools.hmcts.net/jira/browse/PUB-3220

### Change description

Add publishing policy link and page


src/main/controllers/PublishingPolicyController.ts
```diff
@@ -0,0 +1,8 @@
+import { Response } from 'express';
+import { PipRequest } from '../models/request/PipRequest';
+
+export default class PublishingPolicyController {
```
KianKwa, line 4: I think it is worth adding a single unit test for the controller. Even though that is not much to test here, it looks more complete to have tests for every controller :)

src/main/views/publishing-policy.njk
```diff
@@ -0,0 +1,187 @@
+{% extends "template.njk" %}
```
KianKwa, line 1: Can we also add a test to src/test/a11y/tests/public-routes.test.ts?

I think at some stage we might want to see whether we want to continue to use a11y tests for accessibility testing or just use playwright for that. For now please add a test there for completeness.

KianKwa (approved):

